PR #4, opened 2018-11-13
Deal with the case where a certificate has no extensions.

jeamland:


chifflier:
Sorry for the delay, the github mail got in the spam folder

Merged, thanks

Issue #9, opened 2019-07-13
easy way to get the time left?


droundy:
I'm just looking at this crate, and it looks like it will do what I want, but it also looks somewhat awkward to use for a PEM file.  My goal is just do decide if (and when) I need to renew a certificate through lets-encrypt, so I don't need *much* in the way of parsing.

I'm thinking a method on `X509Certificate` or `TbsCertificate` or `Validity` that looks like
```
  fn time_to_expiration(&self) -> Option<std::time::Duration> { ... }
```
would be very convenient to me and presumably to others.  I'd be happy to submit such a patch if you can tell me where you'd like the method to be.  My preference would be at the very top level, but I could imagine putting it in all three or just in `Validity`.

chifflier:
I think the `Validity` structure can be extended with such functions (it seems to be the best place)
Functions could also be added to the certificate itself, so that checking would be simple

chifflier:
BTW, if we use `Duration`, we have to be extra careful, several of its methods can panic (which is not acceptable IMHO, especially in a library crate)

chifflier:
Merged #10, thanks

PR #10, opened 2019-07-14
add time_to_expiration

droundy:
This is an implementation of the idea in issue #9 and looks pretty straightforward.


droundy:
BTW the one hokey thing here is that I wrote a test that assumed it wouldn't take more than 9 seconds.  I think this is not *likely* to cause a problem, but thought I'd mention it in case you worry (or in case travis-ci is amazingly slow).

Issue #8, opened 2019-07-13
easier way to parse/read a PEM file?


droundy:
I'm just looking at this crate, and it looks like it will do what I want, but it also looks somewhat awkward to use for a PEM file.  My goal is just do decide if (and when) I need to renew a certificate through lets-encrypt, so I don't need *much* in the way of parsing.

Your [example](https://docs.rs/x509-parser/0.5.0/x509_parser/pem/index.html) for pem parsing looks pretty doable, but I wonder if would be a problem to simply have a `parse_x509_pem`? I'd also prefer an interface returning a `Result` rather than an `IResult`, maybe even a `read_x509_pem` which reads a file and parses it.

If I wrote such a function, would you accept a pull request? It seems silly to put it in my crate, where no one else will use it.

chifflier:
Hi,
Thanks for your ideas. Here are a few answers:
- adding a function `parse_x509_pem` is pretty straightforward, It just has to map the functions `pem_to_der` and `parse_x509_der`. This would be nice to add
- about reading files, I'd prefer to provide functions using the `Read` trait and let the caller read the file, as usually in other crates
- an `IResult` is actually a `Result` (see [nom documentation](https://docs.rs/nom/4.2.3/nom/type.IResult.html))
- the validity of the certificate can be found in the `tbs_certificate.validity` field, of type [Validity](https://docs.rs/x509-parser/0.5.0/x509_parser/x509/struct.Validity.html). The `not_after` field can be checked for expiration. Maybe a function can be added to to that

I'd gladly discuss and accept PRs

chifflier:
Edit: `parse_x509_pem` might not be so easy to write, since it needs to allocate data (to decode base64). Since the parser borrows a reference, it is not trivial to write one that returns both a new `Vec<u8>` and a certificate borrowing it

droundy:
I just got started and noticed this (before reading your response).  It doesn't seem to me that x509 parsing generally is likely to be so performance-sensitive that zero-copy parsing is essential.  Would you be open to a major change to return a data structure that can be cloned, with no lifetime?

chifflier:
I'm not sure it is even possible currently, because internally there are some `DerObject` which you will not easily be able to deep copy, or make `'static`.

[...]

PR #11, opened 2019-07-16
Switch the pem read method to accept any BufRead + Seek object.

droundy:
This should make it easier to use with files.  Could help address #8.


droundy:
Here is an example that would use the new code to read a PEM file, from my (in progress) lets-encrypt-warp crate:
```
fn time_to_expiration<P: AsRef<std::path::Path>>(p: P) -> Option<std::time::Duration> {
    let file = std::fs::File::open(p).ok()?;
    x509_parser::pem::Pem::read(file).ok()?.0
        .parse_x509().ok()?
        .tbs_certificate.validity.time_to_expiration()
}
```

droundy:
Actually, the above example doesn't work.  I've added a test and doc-test that do work, however.

chifflier:
Indeed, using `Write + Seek` is better - I used `Cursor` originally so it can be used directly on slices, but your approach is better.
The PR is fine, just a minor note: please use `assets/` to store data (like the pem file), and the `Cargo.toml` file should have been updated to include it in package. I'll take care of that.
Thanks for your contribution!

PR #14, opened 2019-09-11
Expose the tbsCertificate byte representation

### LINKED ISSUE #13: Feature request: Expose bytes of encoded tbsCertificate
Hi,

for checking the signature the encoded representation of cbsCertificate is needed. I would suggest to add api to expose this representation.

I think this can be implemented with a nom combinator similar to `recognize`. If you interested in this, I can create a pull request.

jannschu:
Closes #13

Edit: The test case was created using the Python package `cryptography`.

### DIFF
--- a/src/x509_parser.rs
+++ b/src/x509_parser.rs
@@ -247,22 +247,23 @@ fn parse_extensions(i:&[u8]) -> IResult<&[u8],Vec<X509Extension>,BerError> {
 }
 
 
-fn parse_tbs_certificate(i:&[u8]) -> IResult<&[u8],TbsCertificate,BerError> {
-    parse_der_struct!(
+fn parse_tbs_certificate(i: &[u8]) -> IResult<&[u8], TbsCertificate, BerError> {
+    use nom::Offset;
+    let (rem, tbs) = map!(
         i,
-        TAG DerTag::Sequence,
-        version:     parse_version >>
-        serial:      map_opt!(parse_der_integer, |x:DerObject| x.as_biguint()) >>
-        signature:   parse_algorithm_identifier >>
-        issuer:      parse_name >>
-        validity:    parse_validity >>
-        subject:     parse_name >>
-        subject_pki: parse_subject_public_key_info >>
-        issuer_uid:  parse_issuer_unique_id >>
-        subject_uid: parse_subject_unique_id >>
-        extensions:  parse_extensions >>
-        (
-            TbsCertificate{
+        parse_der_struct!(
+            TAG DerTag::Sequence,
+            version:     parse_version >>
+            serial:      map_opt!(parse_der_integer, |x:DerObject| x.as_biguint()) >>
+            signature:   parse_algorithm_identifier >>
+            issuer:      parse_name >>
+            validity:    parse_validity >>
+            subject:     parse_name >>
+            subject_pki: parse_subject_public_key_info >>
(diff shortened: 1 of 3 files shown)


src/x509_parser.rs
```diff
@@ -247,35 +247,63 @@ fn parse_extensions(i:&[u8]) -> IResult<&[u8],Vec<X509Extension>,BerError> {
...
-                subject,
-                subject_pki,
-                issuer_uid,
-                subject_uid,
-                extensions
+fn parse_tbs_certificate(i: &[u8]) -> IResult<&[u8], TbsCertificate, BerError> {
+    use nom::{Offset, Slice, error::ParseError, combinator::map};
+    // A combinator which returns the consumed input and the result.
+    // It is similar to the recognize combinator from nom.
+    fn keep_input<'a, O, E:ParseError<&'a [u8]>, F>(parser:F) -> impl Fn(&'a [u8]) -> IResult<&'a [u8], (&'a [u8], O), E>
```
chifflier: Hum, this seems overly complicated. I like combinators, but I'd clearly prefer simplifying code using something like (in the main function, no need for closures/clones)
```rust
let start = i;
let (rem,mut tbs) = parse_tbs(i)?;
let len = start.offset(rem);
// store i[..len] in tbs
Ok((rem, tbs))
```

This looks much more readable

src/x509_parser.rs
```diff
@@ -247,35 +247,63 @@ fn parse_extensions(i:&[u8]) -> IResult<&[u8],Vec<X509Extension>,BerError> {
...
+            match parser(i) {
+                Ok((i, ret)) => {
+                    let index = input.offset(&i);
+                    Ok((i, (input.slice(..index), ret)))
+                },
+                Err(e) => Err(e),
             }
+        }
+    }
+    static DUMMY: &[u8] = &[];
```
chifflier: This should be directly declared line 296 (additionally, it would be `const` and not `static`)

src/x509_parser.rs
```diff
@@ -247,35 +247,63 @@ fn parse_extensions(i:&[u8]) -> IResult<&[u8],Vec<X509Extension>,BerError> {
...
+    named!(
```
chifflier: I'd prefer to have this in a function doing only the parsing of the `TbsCertificate`, it makes maintenance easier.
Alternatively, as said above this part could just be `let (rem, tbs) = do_parse!( /* previous code unchanged */)?;`

tests/readcert.rs
```diff
@@ -8,6 +8,7 @@ use x509_parser::{parse_subject_public_key_info,parse_x509_der,X509Extension};
 use x509_parser::objects::{nid2obj, Nid};
 
 static IGCA_DER: &'static [u8] = include_bytes!("../assets/IGC_A.der");
+static IGCA_DER_TBS: &'static [u8] = include_bytes!("../assets/IGC_A_tbs");
```
chifflier: This is not required, you can just use offsets in `IGCA_DER`

jannschu:
Thx for the review! :-) I implemented the changes in a new commit. I can squash if you want.

nhynes:
@jannschu and @chifflier here's another take on the same: https://github.com/rusticata/x509-parser/pull/18/commits/9b3f2cab2484dc6121efd12513b4493260c3cbba

chifflier:
Since #19 has been merged, I'm closing this PR
Thanks

PR #16, opened 2019-09-16
Expose tbs bytes

nhynes:
This PR exposes the raw bytes of the `tbsCertificate` so that they can be verified against the signature.

### DIFF
--- a/src/x509_parser.rs
+++ b/src/x509_parser.rs
@@ -247,7 +247,8 @@ fn parse_extensions(i:&[u8]) -> IResult<&[u8],Vec<X509Extension>,BerError> {
 }
 
 
-fn parse_tbs_certificate(i:&[u8]) -> IResult<&[u8],TbsCertificate,BerError> {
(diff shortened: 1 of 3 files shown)


src/x509_parser.rs
```diff
@@ -247,7 +247,8 @@ fn parse_extensions(i:&[u8]) -> IResult<&[u8],Vec<X509Extension>,BerError> {
 }
 
 
-fn parse_tbs_certificate(i:&[u8]) -> IResult<&[u8],TbsCertificate,BerError> {
+fn parse_tbs_certificate(i:&[u8]) -> IResult<&[u8],(TbsCertificate,&[u8]),BerError> {
+    const BER_HEADER_SIZE: usize = 4;
```
chifflier: I'm a bit uncomfortable hard-coding the header size: even if it is canonical (DER), length could be greater (or smaller ?) for big (tiny) certificates. There should be a way to build the size.

PR #19, opened 2019-09-18
Expose tbsCertificate bytes

nhynes:
take 2

The test case offset was determined using the following script:
```python
from Crypto.Util.asn1 import DerSequence
der = open('IGC_A.der', 'rb').read()
cert = DerSequence().decode(der)
tbs = DerSequence().decode(cert[0])
print(len(tbs.payload))
```
with the knowledge that a BER header is 4 bytes.

The signature algo was verified using `ring::signature::ECDSA_P256_SHA256_ASN1`

### DIFF
--- a/src/x509.rs
+++ b/src/x509.rs
@@ -96,6 +96,13 @@ pub struct TbsCertificate<'a> {
     pub issuer_uid: Option<UniqueIdentifier<'a>>,
     pub subject_uid: Option<UniqueIdentifier<'a>>,
     pub extensions: Vec<X509Extension<'a>>,
+    pub(crate) raw: &'a [u8]
+}
+
+impl<'a> AsRef<[u8]> for TbsCertificate<'a> {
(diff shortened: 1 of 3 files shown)


jannschu:
Just a quick note from me: There is already a "impl" block for TbsCertificate in `x509.rs` (I oversaw this one at first, too). Personally, I think `as_ref` is used differently across the standard library (for example for `Option` or arrays) and should be changed to something else.

nhynes:
> `as_ref` is used differently across the standard library 

In this case, it makes the client code very slightly cleaner when used with idiomatic crypto libraries:
```rust
verify(tbs_certificate, signature)
// vs
verify(tbs_certificate.as_bytes(), signature)
```

I'm fine with either way.

chifflier:
Hi,
This was hard to choose, but I'm just thinking that we should start somewhere. This PR is a simple change, which I like, so I merged it.
Thanks @jannschu and @nhynes !

PR #18, opened 2019-09-17
Parse certificate revocation lists

nhynes:
requires https://github.com/rusticata/der-parser/pull/19

PR #22, opened 2019-11-25
Add CRL parser

nhynes:
This is a simple rebase of #18


chifflier:
This looks good, but depends on a new release of `der-parser` which breaks API, so I'd like to wait to see if other breaking changes are also merged as well. The most concerning change is https://github.com/rusticata/der-parser/pull/18, which needs to be decided before a release.
I may release an alpha version of `der-parser` before if that helps.

chifflier:
Hi,
Since the only difference with upstream `der-parser` was the type (bytes vs string), I merged it into stable with modifications so it does not break the API.
I'll prepare a new release soon.
Thanks!

PR #27, opened 2020-02-22
Remove println from parser

kpcyrd:
The latest x509_parser is broken for me since 0285fef (0.6.1).

This patch removes the `println!` from the parse function, please release a new update with this change.


chifflier:
Oops, good catch, thanks!

PR #32, opened 2020-06-26
Make extension parsing more robust

g2p:
Fall back to UnsupportedExtension when an extension parser fails with
Incomplete.
This allows parsing the remaining extensions, instead of silently eating
the rest of the certificate with no indication that the parse was
partial.

Let's encrypt certificates triggered this issue.
I was unable to fix the basicConstraints parsing, but at least with this
I am able to access the whole set of extensions.

### DIFF
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -153,7 +153,7 @@ pub(crate) mod parser {
     use der_parser::{oid::Oid, *};
     use nom::{alt, call, do_parse, eof, exact, many1, opt, take, verify, Err, IResult};
 
-    pub(crate) fn parse_extension<'a>(
+    fn parse_extension0<'a>(
         orig_i: &'a [u8],
         i: &'a [u8],
         oid: &Oid,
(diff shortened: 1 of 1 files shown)


chifflier:
Hi,

If I understand correctly, the real error is an error in the BasicConstraint parsing? If it's possible, I'd be very interested in examples of certificates that failed, and fixing this also.

That said, this leaves the question of how to handle failures when deep-parsing the extensions. Since the global DER structure was already validated, I'd be in favor of being not too strict, and returning `UnsupportedExtension` could be a solution. I'll think about the best solution, so I'll just wait one day or two before merging this PR.

Thanks!

g2p:
Sure! I'll open an issue and attach a certificate for the basicConstraints issue.
(I wanted to anonymize the cert a bit, if I couldn't fix the parser on my own)

chifflier:
Merged, thanks!
This gives a way to know that the extension parser fails, while previously it was silently ignored.

PR #35, opened 2020-07-06
Expand the public API slightly

g2p:
For X509Name, exposes parsing as well as accessing the raw data after parsing.

For parsed certificate extensions, this fixes an unnecessarily short lifetime.

### DIFF
--- a/src/x509.rs
+++ b/src/x509.rs
@@ -48,7 +48,7 @@ impl<'a> X509Extension<'a> {
     }
 
     /// Return the extension type or `None` if the extension is not implemented.
-    pub fn parsed_extension(&self) -> &ParsedExtension {
+    pub fn parsed_extension(&self) -> &ParsedExtension<'a> {
         &self.parsed_extension
     }
 }
--- a/src/x509.rs
+++ b/src/x509.rs
@@ -79,6 +79,14 @@ pub struct AlgorithmIdentifier<'a> {
 #[derive(Debug, PartialEq)]
 pub struct X509Name<'a> {
     pub rdn_seq: Vec<RelativeDistinguishedName<'a>>,
+    pub(crate) raw: &'a [u8],
+}
+
+impl<'a> X509Name<'a> {
+    // Not using the AsRef trait, as that would not give back the full 'a lifetime
+    pub fn as_raw(&self) -> &'a [u8] {
+        self.raw
+    }
 }
 
 impl<'a> fmt::Display for X509Name<'a> {
@@ -502,6 +510,7 @@ mod tests {
                     ],
                 },
             ],
+            raw: &[], // incorrect, but enough for testing
         };
         assert_eq!(
             name.to_string(),
--- a/src/x509_parser.rs
+++ b/src/x509_parser.rs
@@ -51,7 +51,7 @@ fn parse_rdn(i: &[u8]) -> BerResult<RelativeDistinguishedName> {
     .map(|(rem, x)| (rem, x.1))
 }
 
-pub(crate) fn parse_name(i: &[u8]) -> BerResult<X509Name> {
+pub fn parse_name(i: &[u8]) -> BerResult<X509Name> {
     parse_der_struct!(
         i,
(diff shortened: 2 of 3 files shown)


src/x509.rs
```diff
@@ -48,7 +48,7 @@ impl<'a> X509Extension<'a> {
     }
 
     /// Return the extension type or `None` if the extension is not implemented.
-    pub fn parsed_extension(&self) -> &ParsedExtension {
+    pub fn parsed_extension(&self) -> &ParsedExtension<'a> {
```
chifflier, line 51: I'd like to understand - the compiler was inferring a wrong lifetime without the explicit one?

src/x509.rs
```diff
@@ -79,6 +79,14 @@ pub struct AlgorithmIdentifier<'a> {
 #[derive(Debug, PartialEq)]
 pub struct X509Name<'a> {
     pub rdn_seq: Vec<RelativeDistinguishedName<'a>>,
+    pub(crate) raw: &'a [u8],
+}
+
+impl<'a> X509Name<'a> {
+    // Not using the AsRef trait, as that would not give back the full 'a lifetime
+    pub fn as_raw(&self) -> &'a [u8] {
```
chifflier, line 87: - I'm also curious about the comment on lifetimes (which seems similar to the above one)
- Could you share the use case where you needed the raw bytes? I'm asking mostly because, if this is useful for several fields, the `as_raw` method could be added to all parsed components

src/x509_parser.rs
```diff
@@ -51,14 +51,23 @@ fn parse_rdn(i: &[u8]) -> BerResult<RelativeDistinguishedName> {
     .map(|(rem, x)| (rem, x.1))
 }
 
-pub(crate) fn parse_name(i: &[u8]) -> BerResult<X509Name> {
+pub fn parse_name(i: &[u8]) -> BerResult<X509Name> {
```
chifflier, line 54: I agree with making this public, but FYI I think I'll just change to a better name (`parse_x509_name`) by adding a commit

src/x509.rs
g2p: Indeed! Lifetime elision was picking up the lifetime on the &self reference, and not the 'a lifetime on the instance.

My usage goes like this:
I have a long-lived ('a) object that carries the entire DER of the certificate.
A function parses it and returns &str data that is sliced from the same DER.
The &str should be able to live as long as the parent DER, but the X509Extension and ParsedExtension structs are local to the function; having the reference lifetime (function local) distinct from the inner 'a lifetime allows me to return the right &'a str.

src/x509.rs
g2p: Similar to the above; the AsRef trait takes no lifetime parameter, and the as_ref function can only return data with the lifetime on the X509Name; if the parsing is done in an inner function, the inner function won't be able to return slices from that.

My use case is sending OCSP requests and validating the reponses; the raw data is sometimes hashed (building the CertID for the OCSP protocol), it can also be copied (some parts of the request should exactly match the certificates that are being checked), and compared (to check that the response really matches the request).

src/x509_parser.rs
g2p: Good idea!

g2p:
I think having as_raw everywhere would prove useful.
Maybe it should be supported in parse_der_struct?

src/x509.rs
chifflier: Ah, I see now, thank you for the details.
I also encountered this kind of limitations within DER, where the compiler did not apply "transitive" lifetimes to inner objects when returned, but agrees when I give it explicitly - maybe I will be able to fix this now!

chifflier:
> I think having as_raw everywhere would prove useful.
> Maybe it should be supported in parse_der_struct?

I'll see if this can be added easily, but that will be a different issue. Just thinking of rewriting `parse_der_struct` gives me headaches :p

chifflier:
Merged, thanks!

PR #41, opened 2020-07-08
Disable chrono's dependency on the time crate

g2p:
This helps users of the library avoid dependency duplication.


chifflier:
I hadn't noticed that this could be disabled. It's good to have one less dependency.
Thanks!

PR #46, opened 2020-08-07
Fix unnecessary use of pem_to_der in example

nicholasbishop:
The `pem` module documentation was a bit confusing because it gave an
example using `pem_to_der`, then said "Another method to parse the
certificate is to use `pem_to_der`" and gave another example using
`pem_to_der`. Fixed by removing the unnecessary call to `pem_to_der`
in the first example.


chifflier:
Merged, thanks!

PR #47, opened 2020-08-07
Make X509Error and PEMError implement the Error trait

nicholasbishop:
I added a dependency on thiserror to make this easier.


chifflier:
Merged, thanks

PR #48, opened 2020-08-07
Remove a bunch of unused error variants from X509Error

nicholasbishop:


chifflier:
Hi,
This PR has conflicts since merging your other PR (I think this is caused by #47)
Can you rebase it and re-submit ?

chifflier:
Hmm, you probably want to wait before re-sending.
I'm considering changing the internal error handling to actually use these error definitions, and provide more helpful messages than things like `BER invalid tag`.

chifflier:
Ok, so to elaborate on the subject (and for leave notes for future thoughts):

- these error types will be used in the future to avoid generic errors
- but for now, I think they can be removed
- introducing new errors will break parts of the existing API: objects need to all return `X509Error`, but functions using `parse_der_struct` internally will require to be rewritten because the existing macros does not allow different error types. It also requires a new version of `der-parser`
- because of that, it will be done in another release

nicholasbishop:
I went ahead and rebased in case it ends up being useful

chifflier:
Merged, thanks

I've pushed some work on useful error messages in a different branch, which will be merged in the next major release (because it breaks API). For now, let's remove things that are not used.

PR #52, opened 2020-10-12
Make parse_tbs_certificate public

sugargoat:
### Motivation

TbsCertificate::as_ref can be used to get a byte representation of the TbsCertificate, and parse_tbs_certificate successfully parses those bytes. This PR exposes parse_tbs_certificate as a public method so it can be used in this manner.

### DIFF
--- a/src/x509_parser.rs
+++ b/src/x509_parser.rs
@@ -264,7 +264,7 @@ fn get_serial_info(o: DerObject) -> Option<(&[u8], BigUint)> {
     Some((slice, big))
 }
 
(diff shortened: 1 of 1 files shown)


sugargoat:
Closing, as I was able to solve my issue another way. Thanks!

chifflier:
Hi,
Glad to know you found another way. That said, I am still considering making this function public, and in fact maybe all other functions.
While working on other crates (to parse PKCS#7 / CMS for examples), there are lots of objects that could be called directly.

PR #54, opened 2020-10-21
Simplify lifetimes in some places.

hbina:
Signed-off-by: Hanif Bin Ariffin <hanif.ariffin.4326@gmail.com>


chifflier:
Applied, thanks!

PR #55, opened 2020-11-01
Add support for Apple specific extension (WebAuthn/App attestation)

agrinman:
This PR adds support for a particular Apple attestation X509 extension: https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server
(OID 1.2.840.113635.100.8.2)

### DIFF
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -660,7 +660,7 @@ pub(crate) mod parser {
             .as_sequence()
             .or(Err(Err::Error(BerError::BerTypeError)))?
             .first()
-            .ok_or(Err::Error(BerError::BerTypeError))?
+            .ok_or(Err::Error(BerError::ObjectTooShort))?
             .as_slice()
             .or(Err(Err::Error(BerError::BerTypeError)))?;
 
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -35,6 +35,8 @@ pub enum ParsedExtension<'a> {
     InhibitAnyPolicy(InhibitAnyPolicy),
     /// Section 4.2.2.1 of rfc 5280
     AuthorityInfoAccess(AuthorityInfoAccess<'a>),
+    /// https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server (OID 1.2.840.113635.100.8.2)
+    AppleDataSecurityAttestationData(&'a [u8])
 }
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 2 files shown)


src/extensions.rs
```diff
@@ -664,6 +669,26 @@ pub(crate) mod parser {
...
+    fn parse_sequence_single_octet_string<'a>(i: &'a [u8]) -> IResult<&'a [u8], &'a [u8], BerError> {
+        let (rest, obj) = parse_der_with_tag(i, BerTag::Sequence)?;
+        let seq = obj
+            .content
+            .as_sequence()
+            .or(Err(Err::Error(BerError::BerTypeError)))?
+            .first()
+            .ok_or(Err::Error(BerError::BerTypeError))?
+            .as_slice()
+            .or(Err(Err::Error(BerError::BerTypeError)))?;
```
chifflier, line 682: This sequence is a bit complex and fragile (all errors will be the same). There is a function in `der-parser` that does exactly that, and could help rewriting this.
Is this extension a SEQUENCE containing an OCTET STRING? If yes, this code can be largely simplified.
The entire function could become something like:
```rust
    fn parse_sequence_single_octet_string<'a>(
        i: &'a [u8],
    ) -> IResult<&'a [u8], &'a [u8], BerError> {
        parse_ber_sequence_defined_g(|_, input| {
            map_res(parse_ber_octetstring, |x| x.as_slice())(input)
        })(i)
    }
```

I have no certificate with this extension. Could you test the code and update/resubmit your PR?

chifflier: Note: the above code requires to add
```rust
use nom::combinator::map_res;
```

agrinman: Thanks, that's awesome. Will add this shortly!

agrinman: The above code didn't work for me and I am not sure why. I believe it's a sequence containing a single der encoded octet string. I did update the middle error above to indicate the sequence is missing an object...

chifflier: I can't find a clear definition of this extension ... Best I could find is https://developer.apple.com/forums/thread/663118 which lets me think the sequence contains a tagged `[1]` value/

Can you provide some test data? Either the certificate, or an extract of the extension (for ex obtained using `openssl asn1parse`)

cpu:
It looks like this branch hasn't had a push for 4 years and has collected some conflicts with `master` in the meantime. I'm going to close it since it will take some investment by an interested party to get it back into shape for review. 

If someone does come along with an interest please also try to find some test data so that test coverage could be added alongside the feature.

Thanks!

PR #58, opened 2020-11-16
Add support for parsing certification requests

djc:
I'd like to add CSR parsing code to this crate. I'm happy to incorporate any feedback you have.

Depends on https://github.com/rusticata/oid-registry/pull/1 (via a Git dependency for now).

### DIFF
--- a/src/x509_parser.rs
+++ b/src/x509_parser.rs
@@ -127,6 +127,16 @@ fn parse_version(i: &[u8]) -> X509Result<X509Version> {
     }
 }
 
+fn parse_required_version(i: &[u8]) -> X509Result<X509Version> {
+    let (rem, hdr) = ber_read_element_header(i).or(Err(Err::Error(X509Error::InvalidVersion)))?;
+    match hdr.tag {
+        BerTag(0) => {
+            map(parse_ber_u32, X509Version)(rem).or(Err(Err::Error(X509Error::InvalidVersion)))
+        }
+        _ => Ok((&rem[1..], X509Version::V1)),
+    }
+}
+
 fn parse_serial(i: &[u8]) -> X509Result<(&[u8], BigUint)> {
     map_opt(parse_der_integer, get_serial_info)(i).map_err(|_| X509Error::InvalidSerial.into())
 }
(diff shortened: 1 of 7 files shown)


src/x509_parser.rs
```diff
@@ -521,6 +551,84 @@ pub fn parse_algorithm_identifier(i: &[u8]) -> X509Result<AlgorithmIdentifier> {
...
+    for attr in v.into_iter() {
+        if attributes.insert(attr.oid.clone(), attr).is_some() {
+            // duplicate attributes are not allowed
+            return Err(Err::Failure(X509Error::DuplicateAttributes));
+        }
+    }
+    Ok((i, attributes))
+}
+
+fn parse_cri_attributes(i: &[u8]) -> X509Result<HashMap<Oid, X509CriAttribute>> {
```
chifflier, line 568: My test with an openssl CSR failed at `parse_cri_attributes`
The created CSR has no attribute, so input is `a0 00` (empty [0] SET OF Attribute)
Adding a test for empty input solved the problem

src/x509_parser.rs
```diff
@@ -521,6 +551,84 @@ pub fn parse_algorithm_identifier(i: &[u8]) -> X509Result<AlgorithmIdentifier> {
...
+    let (i, hdr) = ber_read_element_header(i).or(Err(Err::Error(X509Error::InvalidAttributes)))?;
```
chifflier, line 569: I think using `parse_ber_tagged_implicit_g` would have been better, as it would ensure only the content of the sequence is parsed.
I will do the update

src/x509_parser.rs
```diff
@@ -521,6 +551,84 @@ pub fn parse_algorithm_identifier(i: &[u8]) -> X509Result<AlgorithmIdentifier> {
...
+    (0..hdr.structured)
+        .into_iter()
```
chifflier, line 571: If I get that correctly, this is equivalent to a `if hdr.structured != 0` statement ?

src/x509_parser.rs
```diff
@@ -521,6 +551,84 @@ pub fn parse_algorithm_identifier(i: &[u8]) -> X509Result<AlgorithmIdentifier> {
...
+            subject,
+            subject_pki,
+            attributes,
+            raw: &start_i[..len],
+        };
+        Ok((i, tbs))
+    })(i)
+}
+
+pub fn parse_x509_csr_der<'a>(i: &'a [u8]) -> X509Result<X509CertificationRequest<'a>> {
```
chifflier, line 618: This function is lacking doc (and tests)

chifflier (approved):
Hi,
Thank you for your contribution! I have a few comments in the PR, but at that point you do not need to change your commits or resubmit a PR.
I'm running some tests, and will add additional patches if needed. I will merge it soon.

src/x509_parser.rs
djc: I guess it might be, I probably don't have the same level of ASN.1 understanding you do. :)

djc: Also note that I have only tested this on CSRs that have exactly one attribute, so this could perhaps benefit from testing with 0 and more-than-1 attributes (though I'm not clear if any other attributes are commonly in use).

src/x509_parser.rs
djc: If you focus on the other stuff, shall I focus on adding some docs and tests?

djc: (Or should I wait on seeing what tests you've already written?)

chifflier:
Merged, with some changes to make peace with clippy
I added one test with a CSR, and fixed the issue with empty attributes list.
Any change or addition can be done using new PRs.
Thanks!

PR #59, opened 2020-11-17
Stop using extern crate where unnecessary

djc:


src/lib.rs
```diff
@@ -139,6 +139,6 @@ mod verify;
 pub use verify::*;
 
 // reexports
-pub extern crate der_parser;
-pub extern crate nom;
-pub extern crate num_bigint;
+pub use der_parser;
+pub use nom;
+pub use num_bigint;
```
chifflier, line 144: The goal of these re-exports was to make sure the caller does not need to add dependencies.
I know `pub extern crate` is supposed to be deprecated, but do these statements offer the same feature ?

djc: Yeah, `pub use` functions the same way.

chifflier:
Merged, thanks!

PR #64, opened 2020-11-17
Use a single impl block per type

djc:
Here's another thing that I found a little confusing, having multiple impl blocks per type. I find this version easier to reason about (but feel free to disagree).


chifflier:
I agree, this is easier to read.
For the record, the code was originally split because I though many functions, use statements and code would have to be marked with `#[cfg(feature = "verify")]` and wanted to factorize that. It appears the amount of code was not that important.

chifflier:
Merged, thanks for your work and ideas!

PR #65, opened 2020-11-17
Improve code locality (fixes #60)

### LINKED ISSUE #60: Consider keeping struct definitions and parsers together
Right now it seems most type definitions live in the `x509` module whereas parsers live in `x509_parser`. It would make the code easier to follow and modify if parsers lived closer to the relevant type definitions, for example in an impl block (as in `fn from_bytes(i: &[u8]) -> X509Result<Self> {}`, for example).

djc:
Here's an attempt at redistributing code in certificate/revocation_list/certification_request modules, leaving shared stuff in x509. Pretty much all parsers live right next to their type definitions, if they have one.

With the exception of the extensions module, I think module sizes are pretty manageable. (extensions was at 935 lines before, so it was already large.)

```
 File                                              Lines         Code     Comments       Blanks
 ./src/error.rs                                      105           82            9           14
 ./src/x509.rs                                       486          347           94           45
 ./src/lib.rs                                        169           41          122            6
 ./src/certificate.rs                                474          288          149           37
 ./src/revocation_list.rs                            288          173           92           23
 ./src/time.rs                                       166          131           18           17
 ./src/pem.rs                                        173           79           84           10
 ./src/certification_request.rs                      159          110           41            8
 ./src/cri_attributes.rs                             139          118            6           15
 ./src/objects.rs                                     67           46           15            6
 ./src/extensions.rs                                1103          868          167           68
```


chifflier:
Awesome, thanks!
I will need some time for the review, I hope later today.
It seems there are also some functions renamed, I'll have to make sure the upgrade path is documented.

djc:
Yeah, I wasn't sure which parts of the public API you wanted to preserve (perhaps temporarily), so I mainly just moved code around.

chifflier:
Update: Overall, I'm OK with the PR. I have added visibility changes, to clarify modules/struct paths, are removed the double export of objects (`pub mod x509` + `pub use x509::*`). Now, only modules are visible, but a `prelude` has been added to be able to import everything easily.
I'm still doing some changes, mostly to make function names consistent.

Note: I also hit an unrelated bug in `rustfmt` while upgrading stable: it now messes the `oid!` macro content :/

All of this is visible in the `djc-coupling` branch (https://github.com/rusticata/x509-parser/tree/djc-coupling). It should be merged in master soon, after some more tests.

djc:
Your additional changes look good to me! If you like, I can rebase #66 on top of master once it's merged.

PR #67, opened 2020-11-18
CSR signature verification should take no arguments (fixes #66)

### LINKED ISSUE #66: Why does X509CertificationRequest::verify_signature take a parameter
In reviewing https://github.com/est31/rcgen/pull/41 I saw that the `verify_signature` function on CSRs taking a parameter, with None being passed in the PR. Fearing that the API witholding some feature that upstream provides, I checked the function definition. Apparently it's only to pass an optional public key to use for verification. For *certificates*, such a param makes sense, as they might be signed by other entities, but CSRs are always self-signed, no? What about removing the parameter from the API before a crates.io release?

cc @djc

djc:


est31 (commented):
Maybe also change the rustdoc of the verify_signature function? It's outdated now :). Also it refers to certificates instead of CSRs :).

djc:
> Maybe also change the rustdoc of the verify_signature function? It's outdated now :). Also it refers to certificates instead of CSRs :).

Fixed. 👍

djc:
The rustfmt failure here appears unrelated.

est31 (approved):

chifflier:
Thanks!
I want to finish working on #60/#65 before reviewing it: a module refactor is in progress, and will likely break your PR

djc:
Rebased.

chifflier:
Merged, thanks!

djc:
Thank you! Do you have any timeline for the next release that will include all of my changes so far?

chifflier:
I intend to release all of this as soon as possible, but I first need to release the stable version of `der-parser`. I just updated the API in 5.0.0-beta2, and this should be close to the final release.
I cannot be precise, but I'd say one or two weeks.

djc:
Great, thanks!

chifflier:
FYI, I've just uploaded 0.9.0 to crates.io

PR #70, opened 2021-02-10
fix: X509Name::iter_state_or_province OID value

0xazure:
Hi! Thanks for writing an maintaining this project, it's made the certificate manipulation work I've had to do for a recent project much easier 😁 .

I hit a slight issue using [`X509Name::iter_state_or_province`](https://github.com/rusticata/x509-parser/blob/0b7e17ef7def4b5d4ed685cb85a30c0a8f689b17/src/x509.rs#L317) today, in that it currently pulls the certificate street address instead of the state or province name as suggested by the method name.

It looks like this mismatch between `X509Name::iter_state_or_province` and its OID value was caused in the switch to `oid-registry` in ddb9ac8287a3284534754c133fc1fd0507ac84f2.

Before the switch `X509Name::iter_state_or_province` used `&OID_ST` which was defined as `2.5.4.8`, but when using values from `oid-registry` it uses the constant `&OID_X509_STREET_ADDRESS` which is defined as `2.5.4.9`.

This corrects the OID value to use the constant `&OID_X509_STATE_OR_PROVINCE_NAME` from `oid-registry` so that it pulls the correct data from the certificate.

### DIFF
--- a/src/x509.rs
+++ b/src/x509.rs
@@ -315,7 +315,7 @@ impl<'a> X509Name<'a> {
 
     /// Return an iterator over the `StateOrProvinceName` attributes of the X.509 Name.
     pub fn iter_state_or_province(&self) -> impl Iterator<Item = &AttributeTypeAndValue> {
-        self.iter_by_oid(&OID_X509_STREET_ADDRESS)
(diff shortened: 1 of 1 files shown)


chifflier:
Good catch, thanks!

PR #77, opened 2021-05-28
Update the version of oid-registry

sasa-tomic:
This enables the change from https://github.com/rusticata/oid-registry/pull/4 in this project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ chrono = { version="0.4", default-features=false, features=["std"] }
 data-encoding = "2.2.1"
 lazy_static = "1.4"
 nom = "6.0"
-oid-registry = { version="0.1.1", features=["crypto", "x509"] }
+oid-registry = { version="0.1.3", features=["crypto", "x509"] }
 rusticata-macros = "3.0"
 ring = { version="0.16", optional=true }
 rustversion = "1.0"


chifflier:
Hi,
Since I updated the patch version (and not minor) in `oid-registry`, this shouldn't be necessary except to prevent using older versions. Crates using `x509-parser` only have to `cargo update` to pick the newer version of `oid-registry`.
Did I miss something?

sasa-tomic:
The version of `oid-registry` is pinned https://github.com/rusticata/x509-parser/blob/2685e2f7f2015774fb1d812f37d527b9c0f6bc36/Cargo.toml#L42 so updating the `oid-registry` didn't get propagated here.

sasa-tomic:
I observed the same thing in our repo, where we're using oid-registry `0.1.1` also, although `0.1.3` is now available.

sasa-tomic:
Any update on this @chifflier ?

SergioBenitez:
> The version of `oid-registry` is pinned

It is _not_ pinned. `0.1.1` means `^0.1.1` which means `0.1.x` where `x > 1`.

sasa-tomic:
@SergioBenitez you are right... as long as there is no `Cargo.lock` file in the project.
And this project does not have one, I presume intentionally.

HOWEVER, our project uses x509-parser and does have a `Cargo.lock` file, and this pins the versions of the dependencies (for reproducibility). So, what can we do?
1. re-add the `x509-parser` dependency, to update the `oid-registry` version
2. force a newer version of the `oid-registry` by changing the `Cargo.toml` file in the `x509-parser`, which is what I'm trying to do here.

If you prefer to not bump up the `oid-registry` version, that's fine. Just let me know please.

chifflier:
> @SergioBenitez you are right... as long as there is no `Cargo.lock` file in the project.
> And this project does not have one, I presume intentionally.

The usual practice is to provide a `Cargo.lock` file for executables, not librarires (see https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html and https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries), so `x509-parser` will not include it.

> 
> HOWEVER, our project uses x509-parser and does have a `Cargo.lock` file, and this pins the versions of the dependencies (for reproducibility). So, what can we do?
> 
>     1. re-add the `x509-parser` dependency, to update the `oid-registry` version
> 
>     2. force a newer version of the `oid-registry` by changing the `Cargo.toml` file in the `x509-parser`, which is what I'm trying to do here.
> 

I will include a version upgrade (I'm just finishing some work on `der-parser` before reviewing all issues/PRs, which should happen really soon).

However, I think you are misusing the lock file. The whole point of having/committing one is exactly to be able to update it (and commit the new version) when upgrading a dependency. This is exactly the reason while lock file exists: libraries usually specify dependencies like `version = "1.2"` (so any patch number is fine) while executables repositories will commit the lock file to indicate the exact version of all (direct and indirect) dependencies that were used for the build.
Note: you can upgrade one dependency selectively:
```shell
$ cargo update -p oid-registry
```

PR #90, opened 2021-07-05
Fix certificate validity period check

SergioBenitez:
According to RFC #5280 4.1.2.5, "the validity period for a certificate is the period of time from notBefore through notAfter, inclusive." At present, `is_valid_at` treats `notAfter` as exclusive. This commit fixes the issue by treating `notAfter` as inclusive.


chifflier:
Applied, thanks!

PR #91, opened 2021-07-05
Expose TbsCertificate::raw, TbsCertificate::raw_serial

SergioBenitez:
`TbsCertificate::raw` is required to verify signatures externally. The bytes passed to the parse function (thus available to the consumer of this API) may not coincide with `TbsCertificate::raw` when parsing consumes only _some_ of the bytes. As such, exposing this field is necessary.

I've also exposed `TbsCertificate::raw_serial`, for parity.


chifflier:
Closing, these fields are already accessible using `x509.tbs_certificate.as_ref()` (for `raw`) and `x509.tbs_certificate.raw_serial()`.
This may require some documentation, though.
Thanks!

PR #95, opened 2021-07-27
Add Ed25519 as an signature algorithm.

### LINKED ISSUE #71: Verification of Ed25519 certificates
Hi! We would like to use `X509Certificate::verify_signature` to verify the signature of a certificate created with OpenSSL that was (self-)signed with [an Ed25519 key](https://docs.rs/openssl/0.10.32/openssl/pkey/struct.PKey.html#method.generate_ed25519).

Currently this fails with an `X509Error::SignatureUnsupportedAlgorithm` because the respective OID (1.3.101.112; see also [RFC 8410](https://tools.ietf.org/html/rfc8410)) is [neither supported by the `x509-parser`](https://docs.rs/x509-parser/0.9.1/src/x509_parser/certificate.rs.html#161-176) nor [by the `oid-registry`](https://github.com/rusticata/oid-registry/blob/master/assets/oid_db.txt).

It seems the underlying `ring` library, however, does support Ed25519 via `ring::signature::ED25519`, which also implements the required `ring::signature::VerificationAlgorithm`, so it seems this is a matter of extending the `oid-registry` and wiring all the existing pieces together.

Would you be interested in supporting (the more and more popular) Ed25519, and if yes, what would be a sensible path forward? For example, is the best/fastest way to submit respective PRs to `oid-registry` and then to `x509-parser`? Thanks!

rappet:
Should fix #71. I did test that using a custom certificate made using the rcgen crate. Sadly my OpenSSL installation can't generate Ed25519 signed certificates.

### DIFF
--- a/src/certificate.rs
+++ b/src/certificate.rs
@@ -133,6 +133,8 @@ impl<'a> X509Certificate<'a> {
                 &signature::ECDSA_P256_SHA256_ASN1
             } else if *signature_alg == OID_SIG_ECDSA_WITH_SHA384 {
                 &signature::ECDSA_P384_SHA384_ASN1
+            } else if *signature_alg == OID_SIG_ED25519 {
+                &signature::ED25519
             } else {
                 return Err(X509Error::SignatureUnsupportedAlgorithm);
             };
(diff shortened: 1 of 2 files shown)


chifflier:
Applied, thanks.

This seems to work, after running some tests here. FYI, I followed https://blog.pinterjann.is/ed25519-certificates.html to generate ED25519 certificates using openssl, and it worked.

PR #99, opened 2021-07-30
Fix new clippy errors

xonatius:
Looks like new release of rust brought in new clippy warnings around references.


xonatius:
FIxed in b703fbc23c45ea7a8324cfa22b8f318b900bd193

PR #100, opened 2021-08-03
Add SubjectPublicKeyInfo::raw field

kpp:
Containing a public key itself.

### DIFF
--- a/src/x509.rs
+++ b/src/x509.rs
@@ -225,12 +225,15 @@ impl<'a> FromDer<'a> for RelativeDistinguishedName<'a> {
 pub struct SubjectPublicKeyInfo<'a> {
     pub algorithm: AlgorithmIdentifier<'a>,
     pub subject_public_key: BitStringObject<'a>,
+    /// A raw unparsed PKIX, ASN.1 DER form (see RFC 5280, Section 4.1).
+    pub raw: &'a [u8],
 }
 
 impl<'a> FromDer<'a> for SubjectPublicKeyInfo<'a> {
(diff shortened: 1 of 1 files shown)


src/x509.rs
```diff
@@ -239,9 +243,12 @@ impl<'a> FromDer<'a> for SubjectPublicKeyInfo<'a> {
                 }
             })(i)
             .or(Err(X509Error::InvalidSPKI))?;
+            let end = i;
+            let raw = &sequence[..2 + begin.len() - end.len()]; // 2 additional bytes for SEQUENCE
```
chifflier, line 247: This line looks clearly suspicious to me:
- the hardcoded `2` is wrong (the DER length could be on 1 byte if < 127, and eventually more than 2 bytes
- additionally, `begin` already points to the sequence *content* (not header), so length is not present
- the comment on the `raw` member seems to indicate that the header is included (otherwise, the result cannot be parsed as a DER object).

Anyway, this is not consistent with other `raw` fields (like in `TbsCertificate`), which include the header.
I'd rather stay consistent with other parsers. Could you update the patch, using the same method as for ex in https://github.com/rusticata/x509-parser/blob/x509-parser-0.9.x/src/certificate.rs#L244-L273 ? (include the header, and use the `start_i.offset(i)` method to obtain the length).
Thanks!

chifflier (changes requested):

src/x509.rs
kpp: I will try

kpp: Done! It worked. If only I knew about `nom::traits::Offset`... =)

chifflier:
Applied, thanks!

PR #104, opened 2021-08-12
Extend the lifetimes exposed on TbsCertificate

g2p:
This makes it possible for me to use the extension accessors introduced in v0.10.


g2p:
(CI errors have to do with the lexical-core dependency and are unrelated)

chifflier:
Applied, thanks!

PR #105, opened 2021-08-17
Fix rustdoc.

jgalenson:
The rustdoc build links to the Validate trait, but that is only available when the validate feature is set.  If rustdoc is run without that feature, it will error due to deny(broken_intra_doc_links).  This changes the comment so that it only links to that trait if the feature is set.


chifflier:
Applied, thanks

PR #114, opened 2022-01-26
Tighten dependencies

g2p:
Ensure things build when resolving dependencies to minimal versions.

Checked with:

    cargo +nightly update -Z minimal-versions; cargo check

There's probably a way to do this in CI, but I haven't looked into it.


chifflier:
Merged in 15549de97f1d72b7aed6b6adfe10ee4a1d6a994c
Thanks!

PR #118, opened 2022-02-21
Fix regression with certificate verification

flavio:
Handle verification of ASN.1 DER-encoded ECDSA signatures using the P-256 curve and SHA-384.

Even though this specific combination is not recommended by `ring`, the `verify_signature` method already supports other not recommended combinations.

Moreover, this used to work until the release 0.12.0, but stopped with the 0.13.0 one.


chifflier:
Good catch, thanks!
Scheduled for the next bugfix release.

PR #119, opened 2022-02-22
Depend on correct version of 'time'

SergioBenitez:
The RFC2822 "well-known" format was added in time v0.3.7. As a result, this crate must depend on at least that version.

...or risk builds that fail because an older version of `time` is pulled in that doesn't yet contain the format.


chifflier:
Applied, thanks!

PR #123, opened 2022-04-15
Update time dependency to 0.3.9.

jasonish:
This is to address the cargo audit issues around time and the RustSec
advisory https://rustsec.org/advisories/RUSTSEC-2020-0071.html.

Migrate from time 0.1 to time 0.3. Due to issues parsing YYMM formatted
years with time 3, provide a crude date parser.

This is for use in the Suricata 6.0.x branch which currently uses
x509-parser 0.6.5.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ base64 = "0.11"
 nom = "5.0"
 rusticata-macros = "2.0.2"
 num-bigint = "0.2"
-time = "0.1"
+time = { version = "~0.3.9", default_features = false, features = ["std"] }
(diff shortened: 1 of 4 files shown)


victorjulien:
Will this bump the MSRV? In suri 6 we use 1.41.1 currently.

jasonish:
It does.  I was working yesterday on the assumption it built on 1.32 or something.  I was looking at the wrong page, and obviously didn't test.

chifflier:
Applied, thanks!
It is a breaking change (some fields are public), however the situation with the `time` crate have been ongoing for too long and justifies a forced update.

PR #129, opened 2022-11-14
Added support for parsing challenge password attribute in CSR's

bkstein:
This branch adds support for parsing a challenge password attribute in a CSR.

Please note: https://github.com/rusticata/oid-registry/pull/10 is a prerequisite, as it adds `OID_PKCS9_CHALLENGE_PASSWORD`. This PR is merged, but not yet released.


bkstein:
A remark @chifflier: I think, the attribute parsing could be improved. Currently, `X509CertificationRequest::from_der()` parses the CSR and knows the (challenge password) attribute's value. This value is held in `X509CriAttribute.parsed_attribute`, which is not visible outside the crate:
```rust
pub struct X509CriAttribute<'a> {
    pub oid: Oid<'a>,
    pub value: &'a [u8],
    pub(crate) parsed_attribute: ParsedCriAttribute<'a>,
}
```
Why is that? A user of the x509-parser crate needs to re-parse `X509CriAttribute.value` instead. I think, the already parsed attribute value should be made available for users. What do you think?

bkstein:
I just compared `CriAttribute` to `X509Extension` and found
```rust
impl<'a> X509Extension<'a> {
    ...
    /// Return the extension type or `UnsupportedExtension` if the extension is not implemented.
    #[inline]
    pub fn parsed_extension(&self) -> &ParsedExtension<'a> {
        &self.parsed_extension
    }
}
```

We could do that in a similar manner for attributes

```rust
impl<'a> CriAttribute<'a> {
    ...
    /// Return the extension type or `UnsupportedAttribute` if the attribute is unknown.
    #[inline]
    pub fn parsed_attribute(&self) -> &ParsedCriAttribute<'a> {
        &self.parsed_attribute
    }
}
```

bkstein:
I will check my proposal and set this request to draft.

bkstein:
Seems to work.

bkstein:
@chifflier The checks will fail until oid-registry with OID for challenge password is released.

chifflier:
> @chifflier The checks will fail until oid-registry with OID for challenge password is released.

oid-registry 0.6.1 has just been released with the required OID

src/cri_attributes.rs
```diff
@@ -99,12 +118,39 @@ pub(crate) mod parser {
...
+    fn parse_extension_request_attr(i: &[u8]) -> X509Result<ParsedCriAttribute> {
         map(
             parse_extension_request,
             ParsedCriAttribute::ExtensionRequest,
         )(i)
     }
+
+    pub(super) fn parse_challenge_password(i: &[u8]) -> X509Result<ChallengePassword> {
+        // I'm sure, there is a more elegant way to try multiple parsers until the first succeeds,
+        // but I don't know nom well enough to implement it.
```
chifflier, line 130: The nom parser for this is [alt](https://docs.rs/nom/latest/nom/branch/fn.alt.html so I supposed the code below can be rewritten as something like `alt(parse_der_utf8string, parse_der_printablestring)`.
Example: https://github.com/rusticata/x509-parser/blob/f6d25f6bbcee178d143476752fb314d03662aa25/src/x509.rs#L153-L155

chifflier (changes requested):
Overall the PR looks fine, thanks!
Can you rebase to the latest master (now that `oid-registry` is released) and address the minor point in comment before merge?

tests/readcsr.rs
```diff
@@ -62,9 +65,35 @@ fn read_csr_with_challenge_password() {
...
+    let challenge_password_attr = csr
+        .certification_request_info
+        .find_attribute(&OID_PKCS9_CHALLENGE_PASSWORD)
+        .expect("Challenge password not found in CSR");
+
+    // 1. Check: Parse value
+    let (rem, challenge_password_from_value) =
+        Set::from_der_and_then(challenge_password_attr.value, |i| {
+            let (i, challenge_password) = String::from_der(i)?;
+            Ok((i, challenge_password))
```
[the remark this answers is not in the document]
bkstein: Yes, I think your right. I'll fix that.

src/cri_attributes.rs
```diff
@@ -99,12 +118,39 @@ pub(crate) mod parser {
...
+            if let Ok((rem, obj)) = parse_der_utf8string(i) {
+                (rem, obj)
+            } else if let Ok((rem, obj)) = parse_der_printablestring(i) {
+                (rem, obj)
+            } else {
+                return Err(Err::Error(X509Error::InvalidAttributes));
+            }
+        };
+        match obj.content {
+            BerObjectContent::PrintableString(s) | BerObjectContent::UTF8String(s) => {
```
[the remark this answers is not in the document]
bkstein: Got it, thanks. However, digging into RFC 2985, 5.2.2 I saw, that all string types should be supported (not only the recommended `PrintableString` and `UTF8String`). I extended the parser with the other types using `nom::branch::alt`.

bkstein:
Changes are implemented. Thanks for reviewing!

src/cri_attributes.rs
```diff
@@ -99,12 +122,43 @@ pub(crate) mod parser {
...
+            parse_der_numericstring,
+            parse_der_bmpstring,
+            visiblestring,
+            parse_der_generalstring,
+            parse_der_objectdescriptor,
+            parse_der_graphicstring,
+            parse_der_t61string,
+            parse_der_videotexstring,
+            parse_der_ia5string,
+        ))(i)
```
chifflier, line 145: Hum, this is not very clear to me (and the new `alt` parser is quite complex - and possibly slow).
I have found `challengePassword` in section 5.4.1 of RFC 2985, defined as `DirectoryString`.
Now, `DirectoryString can be found in RFC 5280:
```
   DirectoryString ::= CHOICE {
         teletexString           TeletexString (SIZE (1..MAX)),
         printableString         PrintableString (SIZE (1..MAX)),
         universalString         UniversalString (SIZE (1..MAX)),
         utf8String              UTF8String (SIZE (1..MAX)),
         bmpString               BMPString (SIZE (1..MAX)) }
```
I don't really see why `ObjectDescriptor`, `NumericString`, etc are listed here.
So my guess is that you probably do not need all these types, but only those from the list. 

You should probably keep the ASN.1 definition in a comment for the function - this helps a lot when trying to interpret the code.

Also, you should order them by "probability" of being found - the `alt` parser will try them one at a time, so the order is important and you should keep `PrintableString` and `UTF8String` parsers at the beginning.

chifflier (changes requested):

src/cri_attributes.rs
bkstein: Correct. I missed the last 3 words in the RFC:
```
   ChallengePassword attribute values generated in accordance with this
   version of this document SHOULD use the PrintableString encoding
   whenever possible.  If internationalization issues make this
   impossible, the UTF8String alternative SHOULD be used.  PKCS #9-
   attribute processing systems MUST be able to recognize and process
   all string types in DirectoryString values.
```
Should be fixed and I kept `UTF8String` and `PrintableString` in the pole position.

chifflier (approved):
Looks great now, thanks for your patience!
I will merge it as soon as possible

PR #130, opened 2022-11-21
Add verify signature method to revocation list

fzpp:
This branch adds a method to verify the signature of revocation list. For this purpose, the code for certificate verification is extracted and reused.


chifflier:
Thanks for the PR
Clippy warnings are not caused by this, so they can be ignored (I will fix them separately)
Looks good to me, I will merge this (with minor fixes) as soon as possible

PR #133, opened 2022-12-27
Fix FromDer of CRLDistributionPoints

duskmoon314:
Implement asn1_rs::FromDer for CRLDistributionPoints via defining a new struct.
Also, implement Deref to avoid breaking.


chifflier:
Applied, thanks!

PR #136, opened 2023-01-16
Switch base64 decoding to `data-encoding` crate

sdroege:
It was already previously used elsewhere in this crate and depending on two base64 implementations doesn't seem ideal.

----

Also fixes https://github.com/rusticata/x509-parser/pull/134

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,6 @@ validate = []
 
 [dependencies]
 asn1-rs = { version = "0.5", features=["datetime"] }
-base64 = "0.21"
 data-encoding = "2.2.1"
 lazy_static = "1.4"
 nom = "7.0"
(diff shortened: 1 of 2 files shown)


chifflier:
Oh, I missed this when merging #138 

This looks interesting, but we'd need to ensure that the features provided are really similar. Do you think this is the case?
If yes, could you rebase and resubmit? Thanks!

sdroege:
It's the same base64 variant, if that's your question :)

https://docs.rs/data-encoding/latest/data_encoding/ has a comparison between `data-encoding`, `base64` and GNU base64 for some other properties.

If this sounds good to you I'll rebase the PR.

chifflier:
Yes, in the past I encountered some weird base 64 variants (some remove the trailing ==, others are said web-safe, etc) and I did not check the RFC to find which is used in certificates. It does not seem to be an exotic one.
I'm fine switching crates, so please go ahead

PR #137, opened 2023-01-30
Use SPDX license format

jbtrystram:
The use of `/` has been deprecated in favor of  `OR`. This causes the cyclonedx cargo extention to fail when generating an SBOM


chifflier:
Applied, thanks!

PR #139, opened 2023-03-02
docs: fix broken FromDer trait link in README.

cpu:
In ac46206 the `traits` module and `traits::FromDer` trait were replaced with a `pub use asn1_rs::FromDer;` directive in the `prelude` module, leaving a broken Rustdoc link in the project `README`.

This commit updates the link target to point to the `prelude` module's trait instead of the now-removed `traits` module's trait. Along the way I also removed some superfluous seeming square brackets around `X509Certificate::from_der` that were being rendered as part of the text.


chifflier:
Applied, thanks!

PR #141, opened 2023-03-22
Fix receiver lifetimes in `AttributeTypeAndValue`.

SergioBenitez:
This builds off https://github.com/rusticata/asn1-rs/pull/22.

### DIFF
--- a/src/x509.rs
+++ b/src/x509.rs
@@ -82,13 +82,13 @@ impl<'a> AttributeTypeAndValue<'a> {
 
     /// Returns the attribute type
     #[inline]
-    pub const fn attr_type(&self) -> &Oid {
+    pub const fn attr_type(&self) -> &Oid<'a> {
         &self.attr_type
     }
 
     /// Returns the attribute value, as `ANY`
     #[inline]
-    pub const fn attr_value(&self) -> &Any {
+    pub const fn attr_value(&self) -> &Any<'a> {
         &self.attr_value
     }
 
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Note that this PR depends on https://github.com/rusticata/asn1-rs/pull/22, so we expect tests to fail until that PR is merged and dependencies are updated as needed.

SergioBenitez:
Checking on this. Can we get this merged?

SergioBenitez:
Checking in again.

src/x509.rs
```diff
@@ -115,7 +115,7 @@ impl<'a> AttributeTypeAndValue<'a> {
 
     /// Get the content as a slice.
     #[inline]
-    pub fn as_slice(&'a self) -> &'a [u8] {
+    pub fn as_slice(&self) -> &'a [u8] {
```
chifflier, line 118: This does not look correct. By forcing only the return to satisfy `'a`, the compiler assigns an unnamed lifetime to `self`, and they do not match. Removing the return lifetime fixes that.
I'll add some commits.

chifflier:
Manually merged in 109e0aa031caeab78b44a71befea7be89452ebd5 and 853d7ee1de14bd57c65e6b6493e9a4655740e490
Thanks!

src/x509.rs
SergioBenitez: This is not the way lifetimes work, and your reversal of this change partially obviates the intended consequences of this PR. Please see my comment in https://github.com/rusticata/asn1-rs/pull/22#issuecomment-1507155046 where I address your misconceptions about lifetimes.

> By forcing only the return to satisfy 'a, the compiler assigns an unnamed lifetime to self, and they do not match.

You seem think that "they do not match" is a problematic statement, but in fact, that's quite literally the point here: we _want_ the receiver's lifetime (`&self`) to differ from the returned lifetime (`&[u8]`). Otherwise, we ask for unification, which means that the receiver must live as long as the return value and vice-versa, even when no such relationship is needed.

Your commit has obviated the meaningful change here and resulted in a strictly worse situation wherein the content of the struct is typed to a particular reference (the receiver, created exactly when the method is called - it doesn't exist otherwise) to the struct. This only makes sense when the struct owns the data, which is not the case here.

SergioBenitez:
Please revert https://github.com/rusticata/x509-parser/commit/853d7ee1de14bd57c65e6b6493e9a4655740e490 and any other such changes and publish a new breaking release.

PR #144, opened 2023-06-18
revocation_list: use correct OID for CRL number.

cpu:
Previously the `CertificateRevocationList`'s `crl_number` fn used the wrong OID when iterating CRL extensions looking for the CRL number extension. It should be using `OID_X509_EXT_CRL_NUMBER` (id-ce 20) but was using `OID_X509_EXT_BASIC_CONSTRAINTS` (id-ce 19).  This results in yielding `None` even for CRLs that do have a CRL number ext.

This commit fixes the mismatched OID and properly yields the CRL number for CRLs with this extension present.


chifflier:
This looks like a typical copy/paste error :)
Applied, thanks!

PR #147, opened 2023-08-26
x509: replace `fold` on `Try` types with `try_fold`.

cpu:
Latest clippy is flagging two instances of using `fold` on a type implementing `Try`, recommending the usage of `try_fold` instead.

This commit fixes both occurrences to resolve the clippy errors. See https://rust-lang.github.io/rust-clippy/master/index.html#/manual_try_fold for more detail.


chifflier:
Hi,
Thank you for your PR. It seems to even simplify the code, so I like it :)

Merged, thanks

PR #152, opened 2024-02-11
Add feature for ring's less-safe-getrandom-custom-or-rdrand

coder0xff:


cpu:
Hi there,

I think like in https://github.com/rustls/rcgen/pull/224 the solution here should be for you to depend on `ring` directly, and set the non-default feature flags you require as appropriate. 

I'm going to close this issue, but if I've misunderstood the problem you're facing (_perhaps you could include some motivation in the PR description?_) we can revisit.

PR #155, opened 2024-03-18
clippy: fix new findings, add cron schedule to CI

cpu:
This branch fixes some `clippy` findings that are breaking CI for new builds ([for example](https://github.com/rusticata/x509-parser/actions/runs/8306288643/job/22734160291?pr=154)).

It also adds `schedule` cron triggers to the `clippy-checks` and `rust` workflows so that we can catch these things sooner when there isn't active development work pushing new branches on a regular cadence.

### DIFF
--- a/src/certificate.rs
+++ b/src/certificate.rs
@@ -15,13 +15,11 @@ use crate::x509::{
 use crate::verify::verify_signature;
 use asn1_rs::{BitString, FromDer, OptTaggedImplicit};
 use core::ops::Deref;
-use der_parser::ber::Tag;
 use der_parser::der::*;
 use der_parser::error::*;
 use der_parser::num_bigint::BigUint;
 use der_parser::*;
 use nom::{Offset, Parser};
-use oid_registry::Oid;
 use oid_registry::*;
 use std::collections::HashMap;
 use time::Duration;
--- a/src/certificate.rs
+++ b/src/certificate.rs
@@ -204,6 +204,12 @@ impl X509CertificateParser {
     }
 }
 
+impl Default for X509CertificateParser {
+    fn default() -> Self {
+        X509CertificateParser::new()
+    }
+}
+
 impl<'a> Parser<&'a [u8], X509Certificate<'a>, X509Error> for X509CertificateParser {
     fn parse(&mut self, input: &'a [u8]) -> IResult<&'a [u8], X509Certificate<'a>, X509Error> {
         parse_der_sequence_defined_g(|i, _| {
@@ -600,6 +606,12 @@ impl TbsCertificateParser {
         TbsCertificateParser {
             deep_parse_extensions,
         }
+    }
+}
+
+impl Default for TbsCertificateParser {
+    fn default() -> Self {
+        TbsCertificateParser::new()
     }
 }
 
--- a/.github/workflows/clippy-check.yml
+++ b/.github/workflows/clippy-check.yml
@@ -1,4 +1,9 @@
-on: push
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 18 * * *'
+
 name: Clippy check
 jobs:
   clippy_check:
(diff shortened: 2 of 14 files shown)


cpu:
I'll try to find some time for some other small CI cleanups in the near future but wanted to get this up to unbreak CI for external contributors. 

Some items on my mind for follow-up: 
* The way clippy integration is being handled seems to [always fail](https://github.com/cpu/x509-parser/actions/workflows/clippy-check.yml) on forked repos. It would be nice to fix that.
* There are a lot of "Node.js 16 actions are deprecated" warnings from the actions-rs tooling. Most of the projects I contribute to have moved away from those to [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) and I suggest we do the same. The [actions-rs organization](https://github.com/actions-rs) has been archived and the tooling is abandonware :-(
* I've had good luck in other repos adding [cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks) and [cargo-check-external-types](https://github.com/awslabs/cargo-check-external-types) to help catch issues. Could be worthwhile here too :+1:

cpu:
@chifflier Can I bug you for a review? :bow:

chifflier:
Hi,
Thank you for your time, and for this PR! I'm currently reviewing it. I also have hit the warnings from clippy-unstable in some other projects and find this "forward-breakage" quite annoying, but in some way it is the exact reason why I added unstable in the CI :)

> I'll try to find some time for some other small CI cleanups in the near future but wanted to get this up to unbreak CI for external contributors.
> 
> Some items on my mind for follow-up:
> 
>     * The way clippy integration is being handled seems to [always fail](https://github.com/cpu/x509-parser/actions/workflows/clippy-check.yml) on forked repos. It would be nice to fix that.

I wasn't aware of that, it is indeed a problem

> 
>     * There are a lot of "Node.js 16 actions are deprecated" warnings from the actions-rs tooling. Most of the projects I contribute to have moved away from those to [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) and I suggest we do the same. The [actions-rs organization](https://github.com/actions-rs) has been archived and the tooling is abandonware :-(

I'm all in favor of a move to a maintained code. Since there are many repositories under the rusticata organization, I would try to apply this to all of them. I don't think we use specific or complex features from this action, so this should not get too complex.
Do you want to work on it? That would  be awesome!

> 
>     * I've had good luck in other repos adding [cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks) and [cargo-check-external-types](https://github.com/awslabs/cargo-check-external-types) to help catch issues. Could be worthwhile here too 👍

I use `cargo-semver-checks` in my local tests before release, but never added it to the CI. That would be nice!
I do not use the other tool so I can't tell, but globally speaking I prefer to add as many checks as possible in the CI.

Thank you again for these great ideas and proposals

src/certificate.rs
```diff
@@ -206,6 +204,12 @@ impl X509CertificateParser {
     }
 }
 
+impl Default for X509CertificateParser {
+    fn default() -> Self {
+        X509CertificateParser::new()
+    }
```
chifflier, line 210: (just a comment) I know clippy prefers the `Default` trait (especially when `new()` takes no argument like here), however it does not allow declaring the method `const` nor `inline` (which I do use a lot).
So, adding an implementation for `Default`, but *not* replacing `new`, looks good to me

chifflier (approved):
Looks good to me

chifflier:
Applied, thanks!

cpu:
Thanks for the feedback :-)

> Do you want to work on it? That would be awesome!

Yes, I will try to open some PRs over the next ~week or so.

cpu:
> Yes, I will try to open some PRs over the next ~week or so.

Here's the first: https://github.com/rusticata/rusticata-macros/pull/8 Let's get that one approved and then I will do the other repos once I know the changes look good.

PR #156, opened 2024-03-28
Add support for RSA-PSS signature verification

nthuemmel-scontain:


assets/rsa-pss/README.md
```diff
@@ -0,0 +1,7 @@
+# Generating Test Certificates
```
cpu, line 1: small nit: WDYT about replacing this `.md` documenting the commands to run with a `.sh` that runs them?

src/verify.rs
```diff
@@ -1,10 +1,13 @@
...
+use crate::signature_algorithm::RsaSsaPssParams;
+use asn1_rs::{Any, BitString};
 use oid_registry::{
-    OID_EC_P256, OID_NIST_EC_P384, OID_PKCS1_SHA1WITHRSA, OID_PKCS1_SHA256WITHRSA,
+    OID_EC_P256, OID_NIST_EC_P384, OID_NIST_HASH_SHA256, OID_NIST_HASH_SHA384,
+    OID_NIST_HASH_SHA512, OID_PKCS1_RSASSAPSS, OID_PKCS1_SHA1WITHRSA, OID_PKCS1_SHA256WITHRSA,
     OID_PKCS1_SHA384WITHRSA, OID_PKCS1_SHA512WITHRSA, OID_SHA1_WITH_RSA, OID_SIG_ECDSA_WITH_SHA256,
     OID_SIG_ECDSA_WITH_SHA384, OID_SIG_ED25519,
 };
+use std::convert::TryFrom;
```
cpu, line 10: I thought this might be superfluous but this crate is still using `edition = "2018"` and requires the import :+1: No action req'd, just leaving a note for the next reviewer.

src/verify.rs
```diff
@@ -76,3 +87,27 @@ fn get_ec_curve_sha(
...
+/// Not all algorithms are supported, we are limited to what `ring` supports.
+/// Notably, the SHA-1 hash algorithm is not supported.
+fn get_rsa_pss_verification_algo(
+    params: &Option<Any>,
+) -> Option<&'static dyn ring::signature::VerificationAlgorithm> {
+    use ring::signature;
+
+    let params = params.as_ref()?;
+    let params = RsaSsaPssParams::try_from(params).ok()?;
+    let hash_algo = params.hash_algorithm_oid();
```
cpu, line 102: optional nit: I think some of the intermediate bindings might not add much value here:

```suggestion
    let hash_algo = RsaSsaPssParams::try_from(params.as_ref()?)
        .ok()?
        .hash_algorithm_oid();
```

cpu (approved):
Thanks! This looks good to me. :rocket: 

I rebased this on https://github.com/rusticata/x509-parser/pull/155 locally and confirmed there weren't any new clippy findings. It builds, lints, and tests green with that branches fixes applied.

src/verify.rs
chifflier: Functionally speaking, they are not required, but I think they makes debugging easier, and sometimes also easier fights with the compiler types engine :)

chifflier:
Applied, thanks!

PR #158, opened 2024-04-12
project CI refresh

cpu:
Continuing the CI standardization and refresh started in https://github.com/rusticata/rusticata-macros/pull/8

Note: I've left out the `cargo check-external-types` CI addition I've been adding in other refresh PRs. Something about this repo is giving the tool some trouble (_it seems to stall in CI after "Examining all public types..."_). I'd like to revisit this when I have more time to debug so for now I've left it out.


chifflier (approved):

chifflier:
Merged, thanks

I do not have ideas yet for the problem with `cargo check-external-types`, but I suspect this is caused by some complexity in using types from multiple crates that end up being the same, but reexported (for ex `asn1-rs::IResult` is the same as `der-parser::IResult` and also as `nom::IResult`).
As explained earlier, they need to be the same to be able to easily combine them. However, reexporting may not be the best solution, but I have not investigated other solutions.

PR #161, opened 2024-07-31
Upgrade `asn1-rs` to version 0.6.2.

plusvic:
This fixes a possible panic while parsing malformed X509 certificates, as described in https://github.com/rusticata/asn1-rs/issues/40

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ verify = ["ring"]
 validate = []
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


chifflier:
Hi,
Thank you for your contribution.

Note: this is not strictly required to fix the possible panic, since it is fixed in the dependency `asn1-rs-derive` (version 0.5.1) and already pulled by `cargo update`.
That said, the proposed change is good to ensure a non-vulnerable version cannot be used.

chifflier (approved):

PR #195, opened 2025-03-12
CI: build with `--locked`, and add one job for stable build without `--locked`

chifflier:
Changes for branch: master

- CI: Build with `--locked`
- CI: Add one job for stable build without `--locked`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -23,6 +23,16 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
       - name: Cargo update
+        run: cargo update --locked
+      - run: RUSTFLAGS="-D warnings" cargo check --locked
+
+  check-notlocked:
+    name: Check (not locked)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Cargo update
         run: cargo update
       - run: RUSTFLAGS="-D warnings" cargo check
 
@@ -41,8 +51,8 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
       - name: Cargo update
-        run: cargo update
-      - run: RUSTFLAGS="-D warnings" cargo check --all-targets --all-features
+        run: cargo update --locked
+      - run: RUSTFLAGS="-D warnings" cargo check --locked --all-targets --all-features
 
   test:
     name: Test Suite
@@ -51,7 +61,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
-      - run: cargo test --all-features
+      - run: cargo test --locked --all-features
 
   test_features:
     name: Test suite (with features)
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -22,6 +22,16 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust }}
+      - name: Cargo update
+        run: cargo update --locked
```
cpu, line 26: I think we wouldn't want to have `cargo update` anywhere except in the `check-notlocked` job. 

My understanding is using `cargo update` with `--locked` like this will start to exit non-zero if there's an update available with a message like "error: the lock file xxxxxx/Cargo.lock needs to be updated but --locked was passed to prevent this" and that's not what we want.

Based on that I think CI is green right now because we just landed an up-to-date `Cargo.lock` but it will start to fail soon.

Am I misunderstanding anything?

chifflier: I wasn't sure, so I tried to look other repositories and it seems they do not run `cargo update`. However, I'm wondering if this is necessary or not (for ex, when installing MSRV, `cargo update` also updates the database of known packages).
I thought it would be harmless - I'm looking at the documentation for an answer

chifflier: Based on [the cargo book](https://doc.rust-lang.org/cargo/commands/cargo-update.html), my understanding is that it is necessary.
It will raise an error if the dependency resolution changed and cause a change in the lock file. If I'm correct, his should not happen except if a dependency was yanked, or changed its own dependencies in an incompatible way.

Thew cargo book also explicitly says ([`cargo update`](https://doc.rust-lang.org/cargo/commands/cargo-update.html) page, doc for option `--locked`):
`It may be used in environments where deterministic builds are desired, such as in CI pipelines.`

cpu: I don't believe it to be necessary (even in the MSRV case) but I'm open to learning more if I'm wrong!

I think the intent is that using `--locked` throughout CI with a checked-in `Cargo.lock` means we build with the exact dependencies versions specified there in a reliable fashion. 

A `cargo update` is only necessary for the new extra CI task that tests that versions newer than those in the checked-in `Cargo.lock` work, and we don't need `cargo update --locked` in CI at all (since it would just verify whether the checked-in `Cargo.lock` matches the newest available versions for deps).

chifflier:
Superseded by #196

.github/workflows/rust.yml
cpu: > Thew cargo book also explicitly says ([cargo update](https://doc.rust-lang.org/cargo/commands/cargo-update.html) page, doc for option --locked):
> It may be used in environments where deterministic builds are desired, such as in CI pipelines.

That's an interesting find, thanks for linking. I have a feeling this isn't the right advice for this particular subcommand (it makes more sense for, e.g., `cargo build --locked`, but not for `update` IMO). I will try and open an upstream PR to suggest different documentation and see if the `cargo` folks agree with me or not.

Issue #192, opened 2025-03-12
CI: build with `--locked`?


chifflier:
CI build on older versions (like https://github.com/rusticata/x509-parser/actions/runs/13806952921 for maintaining stable branch 0.17) can fail, because it does not use `--locked` and thus updates crates to newer versions. In the above link, this fails because of a constant becoming deprecated, and warning treated as errors by us.

In this case, the problem is solved (dependency bumped), but I'm wondering what would be the good thing to do in general.

Is it a good practice to use `--locked` in CI build? Will it cause problems somewhere?
cc @cpu

cpu:
> Is it a good practice to use --locked in CI build? Will it cause problems somewhere?

I think it's a good practice! It's something we do throughout many of the `rustls` repositories for this exact reason.

In terms of problems, I'm not aware of any big ones. I think some projects also like to test a build with the latest versions of all dependencies as a separate CI task but it isn't something I've typically done. Instead we review dependabot/renovate generated PRs to update the lockfile as semver compat releases happen, and if any break CI on the PR run we triage accordingly.

chifflier:
I seem thank you for the reply. IMHO, this could be done in the test matrix, always building with `--locked` and maybe a task without it if we want to be aware of future breaks

cpu:
> always building with --locked and maybe a task without it if we want to be aware of future breaks

Yup, exactly! 

Did you want to submit a PR for this or should I? I can probably find time, but not until this weekend.

chifflier:
I'll send a PR later today or tomorrow and request a review so you'll be notified
Thanks!

PR #196, opened 2025-03-12
CI: build with `--locked`, and add one job for stable build without `--locked`

chifflier:
Supersedes: #195 
Closes: #192


cpu (approved):
Thanks! This version LGTM.

PR #203, opened 2025-04-28
clippy: fix uninlined_format_args findings

cpu:
Applies the result of `cargo clippy --fix` for findings of the form:

```
warning: variables can be used directly in the `format!` string
   --> src/extensions/generalname.rs:105:47
    |
105 |             GeneralName::OtherName(oid, _) => write!(f, "OtherName({}, [...])", oid),
    |                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
    = note: `#[warn(clippy::uninlined_format_args)]` on by default
help: change this to
    |
105 -             GeneralName::OtherName(oid, _) => write!(f, "OtherName({}, [...])", oid),
105 +             GeneralName::OtherName(oid, _) => write!(f, "OtherName({oid}, [...])"),
```

PR #206, opened 2025-05-04
Add value field to ParsedExtension::UnsupportedExtension

### LINKED ISSUE #205: Add support for custom CSR extensions
Currently calling `requested_extensions` on a CSR using a [custom certificate extensions](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2)  returns [`ParsedExtension::UnsupportedExtension`](https://docs.rs/x509-parser/0.17.0/x509_parser/extensions/enum.ParsedExtension.html#variant.UnsupportedExtension). This means that there is no way to parse a user created custom certificate extension.

A simple solution would be adding a `value` field to `ParsedExtension::Unsupported` which would contain the unparsed data. This would allow the user to manually parse the extension data.

jean-airoldie:
This is the simplest solution I could think off. However this is a breaking change, which isn't ideal.

Closes #205.


cpu:
:wave: Thanks for the PR!

> However this is a breaking change, which isn't ideal.

I believe @chifflier is preparing a new semver breaking release, so that might not be a non-starter.

However, I also [left a comment](https://github.com/rusticata/x509-parser/issues/205#issuecomment-2849208502) showing a way to accomplish this without a code change so perhaps it's better to not add new API surface at all?

jean-airoldie:
Your comment addressed my issue, so I will be closing this.

jean-airoldie:
See [my comment](https://github.com/rusticata/x509-parser/issues/205#issuecomment-2849235551).

cpu:
I think the remaining clippy failures are addressed in https://github.com/rusticata/x509-parser/pull/203.

jean-airoldie:
@cpu Alright, I guess I'll merge when that PR eventually gets merged.

jean-airoldie:
While not strictly required, this API is much nicer than using `X509CertificationRequestInfo::attributes` etc. methods because we don't need to manually decode the `OID_PKCS_9_AT_EXTENSION_REQUEST` extension. See [my comment](https://github.com/rusticata/x509-parser/issues/205#issuecomment-2849235551) for more details.

Here's a specific usage example: https://github.com/jean-airoldie/rcgen/blob/30f01893a8002e5c7dd12a4a224ce3f1b70449ba/rcgen/src/csr.rs?pain=#L209-L218

chifflier:
@jean-airoldie Thank you for the PR and the discussion
The breaking PR has been sent in #207 and is currently under discussion. Please wait for this before changing code, since there is a code refactor that will result in conflicts otherwise.

About the extensions and API, I am totally in favor of updating the API if this makes things easier.
Generally speaking, extensions are stored in both unparsed (reference to bytes) and parsed form in internal objects, so it is possible to expose everything.

I also realize that there is no pattern visitor (like for [certificates and CRL](https://docs.rs/x509-parser/latest/x509_parser/visitor/index.html)). Would it be useful for your use case?

jean-airoldie:
> @jean-airoldie Thank you for the PR and the discussion
The breaking PR has been sent in https://github.com/rusticata/x509-parser/pull/207 and is currently under discussion. Please wait for this before changing code, since there is a code refactor that will result in conflicts otherwise.

Alright, I will wait for that then.

> I also realize that there is no pattern visitor (like for [certificates and CRL](https://docs.rs/x509-parser/latest/x509_parser/visitor/index.html)). Would it be useful for your use case?

If the CSR visitor pattern walk method is recursive, meaning it would walk over each `X509Extension` found in  `X509CriAttribute` with `OID_PKCS_9_AT_EXTENSION_REQUEST` (which as far as I understand is already parsed as `ParsedCriAttribute::ExtensionRequest`), then this would indeed also work for my use case. I just want to avoid having to manually parse the `X509CriAttribute` in the case unsupported extensions (in my case prioprietary) are encountered.

chifflier:
> @chifflier
> 
> > @jean-airoldie Thank you for the PR and the discussion
> > The breaking PR has been sent in #207 and is currently under discussion. Please wait for this before changing code, since there is a code refactor that will result in conflicts otherwise.
> 
> Alright, I will wait for that then.

Hi @jean-airoldie 
#207 has been merged and code has now changed quite a lot. In particular, `extensions/mod.rs` has been split into multiple files to make code easier. I know this will cause some conflicts and am sorry for it :cry: 

Can you check how you code can be adapted? Looking at your code, I _think_ it shouldn't be too complicated

> 
> > I also realize that there is no pattern visitor (like for [certificates and CRL](https://docs.rs/x509-parser/latest/x509_parser/visitor/index.html)). Would it be useful for your use case?
> 
> If the CSR visitor pattern walk method is recursive, meaning it would walk over each `X509Extension` found in `X509CriAttribute` with `OID_PKCS_9_AT_EXTENSION_REQUEST` (which as far as I understand is already parsed as `ParsedCriAttribute::ExtensionRequest`), then this would indeed also work for my use case. I just want to avoid having to manually parse the `X509CriAttribute` in the case unsupported extensions (in my case prioprietary) are encountered.

I see. I wonder if this is more a performance problem (avoiding deep-parsing unknown extensions) or an API problem, because if using the visitor the extensions will be parsed anyway, so not improving performance.

jean-airoldie:
@chifflier Alright I rebased into master. I also squashed my commits so I'm gonna need one review to see if CI passes (forgot about that quirk).

I feel that this API is satisfactory, but as far as I understand a visitor API would also do the trick. For my use case, I think the visitor API would result in slightly less code, so I don't feel its necessarily needed.

As a side note, I'm not a big fan of `ans1_rs::Input::as_bytes2` naming. Wouldn't it make sense to simply name it `as_bytes` and the impl nom `AsBytes` traits as `Input::as_bytes(self)` to avoid the stack overflow?

jean-airoldie:
> You can have a look at the visitor API in https://github.com/rusticata/x509-parser/pull/209
You can have access to the raw values, so it could work for you. That said, storing the data in UnsupportedExtension looks good to me.

Alright I'll try that.

> Agreed, the naming is not satisfying, but I had to use a different function name than the traits and I was not very inspired.

Yeah I guess what I meant is I didn't understand why you wanted to use a different name than the trait. If the `nom::AsBytes` trait isn't used explicitely, the name clash doesn't matter. Specificically, I did `sed -i s/as_bytes2/as_bytes/g **/**.rs` in your asn1-rs codebase and everything seems to work find, except compiler warnings about the nom::AsBytes import being unneeded.

chifflier (approved):

chifflier:
Merged, thanks!

PR #207, opened 2025-05-05
Dev/next asn1 rs 0.8 v0.1

chifflier:
Thie is the development branch for the next release.
This PR includes several major changes (some of the breaking changes):
- a refactor of the code base to better split extensions in different files (and avoid a huge `extensions.rs` file)
- `der_parser` dependency entirely removed, now only `asn1-rs` is used for ASN.1 parsing
- upgrade to `nom` 8 and `asn1-rs` 0.8
- the transition also means a few changes in some object field names and methods

I tried to minimize the breaking changes, but there are a few. I am in the process of documenting them (to be pushed in another PR).

I'm opening the PR since it is blocking other commits, so it can be merged as soon as possible.


cpu:
> chifflier requested a review from cpu 

I will take a look this week :rocket:

chifflier:
@cpu I am asking for review if you have some time, but given the amount of changes I do not ask for a detailed review :)
I tried to update all major things, but there are still some work to do. In any case, the code is working and tested on my set of certificates here.

Maybe a quick way to get the visible changes is to look first at the changes in the examples and tests.

Thank you!

Cargo.toml
```diff
@@ -42,18 +42,18 @@ verify = ["ring"]
...
+asn1-rs = { version = "0.8.0", features=["bigint", "datetime"] }
 data-encoding = "2.2.1"
 lazy_static = "1.4"
 nom = "8.0"
-oid-registry = { version="0.8.1", features=["crypto", "x509", "x962"] }
+oid-registry = { version="0.9", features=["crypto", "x509", "x962"] }
 rusticata-macros = "5.0"
 ring = { version="0.17.12", optional=true }
-der-parser = { version = "10.0", features=["bigint"] }
+# der-parser = { version = "10.0", features=["bigint"] }
```
cpu, line 52: It looks like the der-parser imports are all commented out instead of removed. I think it'd be sensible to do a pass across the branch to clean that stuff up before merge. I haven't flagged each instance individually in my review for that reason.

src/lib.rs
```diff
@@ -168,6 +167,9 @@ use certificate::X509Certificate;
 use error::X509Result;
 use revocation_list::CertificateRevocationList;
 
+/// Default maximum object size (2^32)
+pub const MAX_OBJECT_SIZE: usize = 4_294_967_295;
```
cpu, line 172: nit: Isn't this 2^32-1?

2^32 is 4_294_967_296

src/utils.rs
```diff
@@ -1,3 +1,42 @@
...
+///          teletexString           TeletexString (SIZE (1..MAX)),
+///          printableString         PrintableString (SIZE (1..MAX)),
+///          universalString         UniversalString (SIZE (1..MAX)),
+///          utf8String              UTF8String (SIZE (1..MAX)),
+///          bmpString               BMPString (SIZE (1..MAX))
+///    }
+/// </pre>
+#[derive(Debug, PartialEq, Eq, Choice)]
+#[asn1(parse = "DER", encode = "")]
+pub enum DirectoryString<'a> {
```
cpu, line 20: It feels like a bit of an odd choice for `DirectoryString` to end up in a `utils.rs`. I'm generally skeptical of "util" modules since they can sort of end up as a catch-all junkyard over time.

In this case it feels like both `format_serial` and the `DirectoryString` type would be more at home in `x509.rs`. WDYT?

src/certificate.rs
```diff
@@ -321,7 +325,8 @@ pub struct TbsCertificate<'a> {
     pub issuer_uid: Option<UniqueIdentifier>,
     pub subject_uid: Option<UniqueIdentifier>,
     extensions: Vec<X509Extension<'a>>,
-    pub(crate) raw: &'a [u8],
+    /// `raw` is used to verity signature
```
cpu, line 336: typo: verity -> verify

src/cri_attributes.rs
```diff
@@ -1,49 +1,29 @@
...
+///
+/// <pre>
+/// Attribute               ::= SEQUENCE {
+///     type             AttributeType,
+///     values    SET OF AttributeValue }
+///           -- at least one value is required
+/// </pre>
 #[derive(Clone, Debug, PartialEq)]
 pub struct X509CriAttribute<'a> {
+    /// Atribute identifier
```
cpu, line 21: typo: Atribute -> Attribute

cpu (commented):
As you called out, this was a difficult branch to review. Lots of pieces moving all at once :sweat: I've done my best to skim through the commits but in general I don't have strong confidence that it was a meaningful review. Unfortunately I think this update would need to be approached differently in order to be reviewed properly.

If you aren't in a particular rush I'd like to update a few downstream projects before this is merged to get a sense of how they handle the migration (in particular, `rustls/rcgen`, and a private repo I have for an unfinished certificate linter experimenting). WDYT?

chifflier:
> As you called out, this was a difficult branch to review. Lots of pieces moving all at once 😓 I've done my best to skim through the commits but in general I don't have strong confidence that it was a meaningful review. Unfortunately I think this update would need to be approached differently in order to be reviewed properly.

I know it's a huge diff :/ I'll try to help where possible. The transition to `asn1-rs` 0.8 and the removal of `der-parser` caused a few breaking changes, I am currently listing them in upgrade notes.
Other changes should be easier to follow, in particular the huge file `extensions/mod.rs` was split into per-extension file.

> 
> If you aren't in a particular rush I'd like to update a few downstream projects before this is merged to get a sense of how they handle the migration (in particular, `rustls/rcgen`, and a private repo I have for an unfinished certificate linter experimenting). WDYT? I suspect I can get that finished this upcoming week.

I am in no hurry, and the goal is exactly to have some time to prepare a proper transition, so I'm fine with that. That said, I don't know if there is a good (git or github) way to mark `master` as frozen (to avoid people working on PR that will conflict).

I will push a few other patches these days to clean up some comments and start adding upgrade notes.

Cargo.toml
chifflier: I will add a patch to remove all remains (there are a few comments in files)

src/lib.rs
chifflier: Indeed
Note: this constant is not used anymore (`asn1-rs` takes care of reading objects), so I'm considering removing it.

chifflier:
Note: the transition to `asn1-rs` 0.8 allowed to use some custom derive attributes (like `Sequence` and `Choice`), which greatly simplified some structs.
This also helped better defining when IMPLICIT/EXPLICIT tags are used, and uncover some horrors from ASN.1 specifications: when IMPLICIT tags are used but the inner type is an untagged choice or open type, then the tag is "automagically" converted to EXPLICIT. If only the specifications said that clearly ...

src/certificate.rs
chifflier: Fixed in [605addb](https://github.com/rusticata/x509-parser/commit/605addbe57aa0b95e037cec0d2a7b5e9d75e0f10)

src/utils.rs
chifflier: I agree on the "catch-all junkyard" feeling :)
I'll move these to `x509.rs`. Also, `format_serial` could be moved to an implementation of `LowerHex`.

cpu: > This makes me wonder if we should add a Serial object to encapsulate the raw bytes

That sounds reasonable to me!

cpu:
> or like too much work? 

It might have been better if `main` didn't already have unreleased breaking changes. Perhaps not worth publishing those solo as 0.18.

One possible alternative/tweak is to abandon those (or rather, consider them rolled into this branch) and use `rel-0.17` as the stable pre-asn1-0.8 release branch to try and give projects using x509-parser something stable while looking at the breaking update to this branch. When this merges the release we eventually cut becomes 0.18.

chifflier:
> How about we do something like:
[...]

I can think of a slightly different (and I think easier) plan:

1. Create `x509-parser-0.18` starting from current `master:HEAD` (a92bbab5cdc630f7d2c0410736bee55f0e3710e8)
2. clippy problems can be resolved in the `x509-parser-0.18` branch and a release can be published (there are also a few minor changes in history, but I see no problem to release this)
3. Merge this PR in `master` as soon as possible. This requires only a trivial action since the long commit stacks shares the same history, so no merge commit nor conflicts
4. In `master`, update version to become `0.19.0-beta.1`
5. Continue working as usual

This could be done quite quickly and allow us to work back on `master` almost directly. What do you think?

cpu:
@chifflier That sounds good to me :rocket:

chifflier:
> @chifflier That sounds good to me 🚀

Thank you again! Note that if you have any comment on the code (now, or when upgrading your code) I'd be happy to hear them. I added `UPGRADING.md` with some notes on the most important changes.

cpu:
> If you aren't in a particular rush I'd like to update a few downstream projects before this is merged to get a sense of how they handle the migration (in particular, rustls/rcgen, and a private repo I have for an unfinished certificate linter experimenting).

Quick update on this front: I have `rcgen` building w/ the latest `x509-parser` in a messy WIP branch. I think I'll have some notes to contribute back to the update doc based on the experience. It's not quite ready to go: there are a few failing unit tests I need to sort out when time/energy permit.

PR #210, opened 2025-05-19
remove lazy_static dependency

2ndDerivative:
Lazy primitives are now available in std, let's use them.

This is breaking the MSRV, but when this crate bumps it's done.


chifflier:
Hi,
Quick update: unfortunately, this breaks the MSRV to version 1.80, which will break some crate users.
Since this change is  not really essential (but interesting), I will delay the merge for some time (until MSRV is updated).
Thank you

chifflier:
Re-considering this PR (now the MSRV has been bumped to 1.85)
I had to merge changes manually because of diverging changes since this PR, but all of this is now merged in [16a6621](https://github.com/rusticata/x509-parser/commit/16a662183be9f6dbdafdc4979bdc5abd39b6e46e), so closing.
Thank you

PR #211, opened 2025-07-31
Fix clippy (nightly) warnings: hiding a lifetime that's elided elsewhere is confusing

chifflier:


src/lib.rs
```diff
@@ -186,7 +186,7 @@ pub fn parse_x509_certificate(i: &[u8]) -> X509Result<X509Certificate> {
 /// This function is an alias to [CertificateRevocationList::from_der](revocation_list::CertificateRevocationList::from_der). See this function
 /// for more information.
 #[inline]
-pub fn parse_x509_crl(i: &[u8]) -> X509Result<CertificateRevocationList> {
+pub fn parse_x509_crl<'i>(i: &'i [u8]) -> X509Result<'i, CertificateRevocationList<'i>> {
     CertificateRevocationList::from_der(i)
 }
```
cpu, line 192: Does this need to be a named lifetime? It seems like it could be elided like the est.

```suggestion
pub fn parse_x509_crl(i: &[u8]) -> X509Result<'_, CertificateRevocationList<'_>> {
    CertificateRevocationList::from_der(i)
}
```

src/lib.rs
```diff
@@ -196,7 +196,7 @@ pub fn parse_x509_crl(i: &[u8]) -> X509Result<CertificateRevocationList> {
     note = "please use `parse_x509_certificate` or `X509Certificate::from_der` instead"
 )]
 #[inline]
-pub fn parse_x509_der(i: &[u8]) -> X509Result<X509Certificate> {
```
cpu, line 199: ditto here

src/lib.rs
```diff
@@ -207,6 +207,6 @@ pub fn parse_x509_der(i: &[u8]) -> X509Result<X509Certificate> {
     note = "please use `parse_x509_crl` or `CertificateRevocationList::from_der` instead"
 )]
 #[inline]
-pub fn parse_crl_der(i: &[u8]) -> X509Result<CertificateRevocationList> {
+pub fn parse_crl_der<'i>(i: &'i [u8]) -> X509Result<'i, CertificateRevocationList<'i>> {
```
cpu, line 210: and here

tests/readcsr.rs
```diff
@@ -155,7 +155,7 @@ fn read_csr_with_custom_extension() {
                 assert_eq!(req.extensions.len(), 1);
                 let extension = req.extensions.first().unwrap();
                 assert_eq!(extension.oid, OID_CUSTOM_EXTENSION);
-                assert_eq!(extension.critical, false);
+                assert!(!extension.critical);
```
cpu, line 158: I think the two `assert_eq!` fixes in this file are sensible, but it'd be nicer for review purposes if it was separated into its own commit since it's unrelated to the lifetime elision changes.

cpu (approved):
Thanks! LGTM. Just had a couple nits to consider.

src/lib.rs
chifflier: I has the same question :) I have no real preference, I just found strange to use anonymous lifetimes while the lifetime is known to be exactly the same as input. So, no real preference here

tests/readcsr.rs
chifflier: Haha, my opinion was to think of this as minor, but it's true it doesn't match commit title. I'll split the patch

chifflier:
Superseded by #212

PR #213, opened 2025-07-31
Dev/add aws lc rs v0.1

### LINKED ISSUE #204: Add `aws-lc-rs` feature
The `verifier` feature uses `ring` for cryptographic functions.
The `ring` is in [low maintenance mode](https://github.com/briansmith/ring/discussions/2414), meaning it won't get any new features in the foreseeable future, for instance post-quantum ML-KEM support.

The ecosystem seems to bend towards `aws-lc-rs` for now.

It would be nice if `x509-parser` supports using this `aws-lc-rs` as `cryptographic` provider.

It would help with codebase auditing to ensure `ring` is no longer in the dependency tree

chifflier:
Add a new feature `verify-aws`, providing the same as `verify`, but based on `aws-lc-rs` instead of `ring`.
Closes #204 

Note: to avoid problems having possible mutually exclusive features and provide easier compatibility with previous code, it is possible to activate both features at the same time. In that case, `verify-aws` will be used (and not `ring`).

**important**: this PR is still marked as draft, for discussion in #204. If if is satisfying, it can be converted to a PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,17 +38,19 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 default = []
+verify-aws = ["aws-lc-rs", "verify"]
 verify = ["ring"]
 validate = []
 
 [dependencies]
+aws-lc-rs = { version = "1.0", optional = true }
 asn1-rs = { version = "0.8.0-beta.1", features=["bigint", "datetime"] }
 data-encoding = "2.2.1"
 lazy_static = "1.4"
 nom = "8.0"
 oid-registry = { version="0.9.0-beta.1", features=["crypto", "x509", "x962"] }
-rusticata-macros = "5.0"
 ring = { version="0.17.12", optional=true }
+rusticata-macros = "5.0"
 thiserror = "2.0"
 time = { version="0.3.35", features=["formatting"] }
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -70,6 +70,7 @@ jobs:
           - --features=default
           - --all-features
           - --features=verify
+          - --features=verify-aws
           - --features=validate
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -38,17 +38,19 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 default = []
+verify-aws = ["aws-lc-rs", "verify"]
```
soundofspace, line 41: While this will use `aws-lc-rs` it will also pull in `ring` which might be unwanted. So you probably want to keep these separate and use something like this instead
```
#[cfg(any(feature = "verify", feature = "verify-aws"))]
mod verify;

#[cfg(all(feature = "verify", not(feature = "verify-aws")))]
use aws_lc_rs::signature;

#[cfg(feature = "verify-aws")]
use aws_lc_rs::signature;
```

chifflier: Yeah, I was just doing a quick hack because I need all these cases to work correctly:
- `--features verify`
- `--features verify-aws`
- `--all-features`

I have to check, but if this provides a cleaner that would be great!

cpu:
> I updated the code and submitted a new PR

@chifflier Apologies for the meta-question, but is there a reason you prefer to open a 2nd PR instead of pushing updates based on the feedback to this one? (Either as new commits to be squash-merged when ready, or as a force push replacing the prev contents). 

I noticed you did that with the clippy PR as well. If you're open to changing styles I think keeping everything in one PR as much as possible makes it easier to track the evolution of the work & all the previous rounds of feedback in one place.

chifflier:
> @chifflier Apologies for the meta-question, but is there a reason you prefer to open a 2nd PR instead of pushing updates based on the feedback to this one? (Either as new commits to be squash-merged when ready, or as a force push replacing the prev contents).

I'm doing that because that's usual in many other open-source projects I contribute to (like suricata). This allows different reviews and help tracing the changes between each PR. It also helps keeping comments related to the real files they were related to.
Note: I'm not saying it's the best process, but a habit which also has its advantages.
As a simple rule, I'm not against pushing more commits, but I try to avoid rewriting the history (even in a PR).

PR #214, opened 2025-08-01
Dev/add aws lc rs v0.2

### LINKED ISSUE #204: Add `aws-lc-rs` feature

chifflier:
Replaces #213

Add a new feature `verify-aws`, providing the same as `verify`, but based on `aws-lc-rs` instead of `ring`.
Closes #204 

Note: to avoid problems having possible mutually exclusive features and provide easier compatibility with previous code, it is possible to activate both features at the same time. In that case, `verify-aws` will be used (and not `ring`).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -105,6 +105,13 @@
 //! }
 //! ```
 //!
+//! - The `verify-aws` feature offers the same support for signature verification, but based on
+//!   `aws-lc-rs` instead of `ring`.
+//!
+//! - _Note_: if both `verify` and `verify-aws` features are enabled (which happens when using
+//!   `--all-features`), the verification will use `aws-lc-rs`. It also has the side-effect of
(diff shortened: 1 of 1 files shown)


chifflier:
Hi @cpu , this PR is now ready for review. If you find some time, could you have a look? The plan is also to backport it to 0.18 and release 0.18.0 as soon as possible. Thanks!

PR #218, opened 2025-08-13
Dev 0.18/add raw bytes full struct v0.1

### LINKED ISSUE #217: Expose source bytes for `X509Certificate`
Right now `TbsCertificate` has a private `raw` field that is used to implement `AsRef<[u8]>` to get the original DER-encoded bytes back. This is almost usable, but because it's at the `TbsCertificate` level instead of the `X509Certificate` level it means it doesn't include the signatureAlgorithm or signatureValue, and so it doesn't quite do what I need.

I'd really like the `X509Certificate` object itself to have a private `raw` field so it can expose the original DER-encoded bytes for the whole certificate (perhaps through another `AsRef<[u8]>` impl, or perhaps through a method). My use-case for this is I need to use a separate crate that implements its own certificate parser in order to do certificate path validation, but I use `x509-parser` for everything else, and it's very awkward to try and track the original source bytes for the certificate myself (e.g. right now I parse a p12 file into a set of certificates and then sort them into an ordered chain, and so I've already lost track of which `Vec<u8>` a given cert came from).

Note that using `AsRef<[u8]>` to provide the raw contents changes the lifetime of the returned bytes, it would probably be a good idea to expose an actual method for this so the returned bytes can have the original lifetime.

chifflier:
*target branch is 0.18*

Add `as_raw` method to top-level structs (`X509Certificate`, `CertificateRevocationList` and `X509CertificationRequest`) to expose bytes that were used during parsing.

closes #217

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,9 +111,9 @@ checksum = "1b8e56985ec62d17e9c1001dc89c88ecd7dc08e47eba5ec7c29c7b5eeecde967"
 
 [[package]]
 name = "cc"
-version = "1.2.31"
+version = "1.2.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c3a42d84bb6b69d3a8b3eaacf0d88f179e1929695e1ad012b6cf64d9caaa5fd2"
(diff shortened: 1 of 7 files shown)


cpu (approved):
I had some feedback for the documentation, but the mechanics seem good. Sorry about the slow review time.

PR #220, opened 2025-09-01
fix(verify): also enable all functions when using verify-aws

jalil-salame:
Otherwise you are missing part of the API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,14 +83,14 @@ See also `examples/print-cert.rs`.
 
 # Features
 
-- The `verify` feature adds support for (cryptographic) signature verification, based on `ring`.
+- The `verify` and `verify-aws` features adds support for (cryptographic) signature verification, based on `ring` or `aws-lc` respectively.
   It adds the
   [`X509Certificate::verify_signature()`](https://docs.rs/x509-parser/latest/x509_parser/certificate/struct.X509Certificate.html#method.verify_signature)
   to `X509Certificate`.
 
 ```rust
 /// Cryptographic signature verification: returns true if certificate was signed by issuer
-#[cfg(feature = "verify")]
+#[cfg(any(feature = "verify", feature = "verify-aws"))]
 pub fn check_signature(cert: &X509Certificate<'_>, issuer: &X509Certificate<'_>) -> bool {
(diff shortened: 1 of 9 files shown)


chifflier (approved):

chifflier:
Applied, thanks.
I'll backport this for branch 0.18 as wel

cpu:
Is there a way CI could be updated with the coverage to catch a mixup like this in the future?

chifflier:
> Is there a way CI could be updated with the coverage to catch a mixup like this in the future?

I think it's possible, though I never looked at integrating github CI and coverage. If anyone has pointers, I'm interested :)
In this particular case, it's not that easy to catch, because it depends on a certain set of features ... Note that unit tests were passing, so I (naively) thought the test included this code.

cpu:
> though I never looked at integrating github CI and coverage.

Sorry, I should have been clearer. By "coverage" I just meant that something in CI should fail if existing unit tests aren't working with a particular feature flag combination, not necessarily anything to do with measuring code coverage from the tests.

> In this particular case, it's not that easy to catch, because it depends on a certain set of features

One option is to use [`cargo hack`](https://github.com/taiki-e/cargo-hack)'s ability to test the crate feature powerset. We've had good luck with it in [rustls](https://github.com/rustls/rustls/blob/a84d3e47388f0aa57ed96a81d4f85a46ffdab439/.github/workflows/daily-tests.yml#L139-L166). Depending on how much CI runtime it adds we might need to fiddle with the config like we did there to keep it reasonable.

I don't have time to tackle this myself in the next week or two but would maybe worth thinking about on a rainy day.

jalil-salame:
Note that `cargo hack` wouldn't have caught this since the tests were turned off by the missing feature.

cpu:
> Note that cargo hack wouldn't have caught this since the tests were turned off by the missing feature.

Ah, great point :face_exhaling: Maybe still worth pursuing, but I think I'm convinced this case would have been tricky to catch outside of eyeballing it in review.

jalil-salame:
Yeah... I'd love to have feature aliases like `verify.alias = "any(verify-ring, verify-aws)" `. You could do this with a breaking change and a build script, but that's annoying...

PR #221, opened 2025-09-01
fix(verify): also enable all functions when using verify-aws

chifflier:
Backport fix for feature `verify-aws` (see #220) for branch 0.18


cpu (approved):
Thanks!

PR #223, opened 2025-09-08
Dev/update version and notes

### LINKED ISSUE #222: UPGRADING file does not seem to apply
The version 0.18.0 was released last week. However, the changelog was not updated with the most recent released changes.

I found the UPGRADING file but the changes described there don't seem to have happened, at least the ones I use/checked.

chifflier:
Misc changes to prepare first beta release (which will be 0.19.0-beta.1):
- Fix version number in CHANGELOG and UPGRADING files (closes #222)
- Update version to 0.19.0-beta.1
- chore(cargo): update dependencies `time` and `thiserror`
  - note that `time` depends on `time-core` which unfortunately requires rust 1.81, which explains the next commits
  - compatibility with our current MSRV is still possible by using a `--precise` flag with `cargo update`
- Clarify MSRV policy in README
- switch to `cargo rdme`, add CI checks and update links
  - this helps maintaining `src/lib.rs` and README.md in sync


chifflier:
@cpu ping? I do not mean to hurry. I totally understand if you do not have time, just tell me :)

cpu:
Thanks for the ping and sorry for letting it sit without a reply! I'm a bit more over-subscribed than usual this week and probably won't have a chance to look at it until ~next. You don't have to wait for me of course if that's too long to wait.

dns2utf8:
Quick question, is the code in this PR what landed in the 0.18 release?

Thank you all

chifflier:
Hi @cpu,
Gentle ping for this PR review

> Quick question, is the code in this PR what landed in the 0.18 release?
Yes (and more). This is mostly merging the entry for 0.18.0 in `CHANGELOG.md`, and then bumping version for `master` (next release planned is 0.19.0-beta.1)

CHANGELOG.md
```diff
@@ -6,6 +6,32 @@
...
 ### Thanks
 
+## 0.18.0
+
+### Added/Changed
+
+- Update lock file and dependencies
+- Fix clippy warnings
+- Visitor: add method to visit unknown extension and those with parse errors
+- Add new feature `verify-aws` to used `aws-lc-rs` as crypto provider instead of `ring`
```
cpu: typo: "to used" -> "to use"

CHANGELOG.md
```diff
@@ -6,6 +6,32 @@
...
+  - The features are exclusive, so only one should be used
+  - If both are specified, `aws-lc-rs` is used (but both dependencies are included)
+- Add `as_raw` methods to `X509Certificate`, `CertificateRevocationList` and `X509CertificationRequest`
+  - This method exposes the raw ASN.1 DER bytes used to build the object (#217)
+
+Extensions:
+- Add support for SubjectInfoAccess extension
```
cpu: nit: put "SubjectInfoAccess" in backticks like the other technical terms

CHANGELOG.md
```diff
@@ -6,6 +6,32 @@
...
+- GeneralName: add a new variant `Invalid` so an invalid entry does not stop
```
cpu: nit: ditto here for "GeneralName" in backticks.

CHANGELOG.md
```diff
@@ -6,6 +6,32 @@
...
+  parsing for the entire list of names (for ex in SAN)
+
+### Fixed
+
+- PEM: ignore lines in comments which contain invalid UTF-8 characters (#180)
+
+### Thanks
+
+- Daniel McCarney
```
cpu: Some other folks for consideration:

* `lillyball` for https://github.com/rusticata/x509-parser/issues/217
* `soundofspace` and `stormshield-gt` for feedback on https://github.com/rusticata/x509-parser/pull/213 and https://github.com/rusticata/x509-parser/issues/204

README.md
```diff
@@ -111,6 +111,16 @@ dependencies and for proc-macro attributes support.
 [RFC5280]: https://tools.ietf.org/html/rfc5280
 <!-- cargo-sync-readme end -->
 
+## MSRV policy
+
+This projects tries to maintain compatibility with older version of the rust compiler for the following
```
cpu: nit: typo "with older version" -> "with older version**s**" (plural)

README.md
```diff
@@ -111,6 +111,16 @@ dependencies and for proc-macro attributes support.
...
+durations:
+- `master` branch: _12 months_ minimum
+- older releases: about 24 months
+
+However, due to dependencies and the fact that some crate writers tend to require very recent
+versions of the compiler, this can prove to be difficult. These numbers are given as _best-effort_.
```
cpu, line 141: It might also be worth touching on whether MSRV changes are considered breaking for semver (I would encourage saying that they aren't). Here's the [rustls MSRV policy](https://github.com/rustls/rustls/blob/main/SECURITY.md#minimum-supported-rust-version) as one point of inspiration.

README.md
```diff
@@ -8,6 +6,8 @@
 [![Github CI](https://github.com/rusticata/x509-parser/workflows/Continuous%20integration/badge.svg)](https://github.com/rusticata/x509-parser/actions)
 [![Minimum rustc version](https://img.shields.io/badge/rustc-1.67.1+-lightgray.svg)](#rust-version-requirements)
 
+<!-- cargo-rdme start -->
```
cpu, line 9: Unfortunately it seems like this commit is breaking many markdown links in the rendered preview (https://github.com/rusticata/x509-parser/blob/fe005007af027b0cbf24f8a8bd8fadac204df527/README.md)

I think the link targets aren't quite right.

cpu (commented):
Thanks for the ping and apologies for letting this fall off my radar! Here's some initial feedback.

README.md
cpu: (Would also be nice to get https://github.com/rusticata/x509-parser/pull/230 merged)

README.md
chifflier: Thank you for the link. I mentioned MSRV changes (which we do not consider breaking for semver) and the fact we try to not make MSRV changes in stable/security patches.
However, this proves to be incredibly difficult because of dependencies. For example, `time` tends to bump MSRV every time they fix an issue (like RUSTSEC-2026-0009) :/

README.md
chifflier: This is an annoying limitation of `cargo rdme`: not all intradoc links are supported, for ex ``[`visitor`]`` is not replaced.
I found a workaround by giving the full path ``[`X509Certificate`](crate::certificate::X509Certificate)`` but now `cargo doc` is complaining that the path is redundant
If I push a commit to use this workaround, this will add warnings when running `cargo doc` (and cannot fix all links).

The last solution is to strip the intradoc links when generating the README.

chifflier: About #230 , I cherry-picked the commit in this branch with a few changes to resolve the conflicts and change `src/lib.rs` as well as `README.md`.

README.md
cpu: > However, this proves to be incredibly difficult because of dependencies.

Agreed. It's a tricky balancing act.

> For example, time tends to bump MSRV every time they fix an issue (like RUSTSEC-2026-0009) :/

Yeah, I've felt the pain of this one across a few projects I help maintain. In case it's helpful context, for some we've marked the finding ignored in the cargo audit tooling and fixed the `time` version we test with our MSRV like you've done here. I think that's very reasonable for `x509-parser`. In others, where there is a binary component shipped we sucked up the 1.88 MSRV increase and figured we'd ride the wave of other projects doing the same.

README.md
cpu: > This is an annoying limitation of cargo rdme: not all intradoc links are supported, for ex [`visitor`] is not replaced.

Bummer :-(

Would you be open to a more direct/dumber solution? We use a [small script](https://github.com/rustls/rustls/blob/main/admin/pull-readme) in `rustls` that slurps the header rustdoc from `lib.rs` and inserts it into the README at a fixed location. [In CI](https://github.com/rustls/rustls/blob/714d12078b79b643df2fe18236eb5997bb1b4697/.github/workflows/build.yml#L284-L288) we run the same script and then error if there's a diff to make sure we always update `lib.rs` and not the copy in `README.md`. 

I don't think we've had any significant trouble/friction with this approach but it's not quite as polished as a dedicated tool like `cargo rdme`.

> If I push a commit to use this workaround, this will add warnings when running cargo doc (and cannot fix all links).

I wonder if it's possible to silence/allow just that one warning about redundant paths. That seems preferable to allowing all warnings, or stripping links if you want to stick with `cargo rdme`.

chifflier: Hi @cpu 
I am trying to resurrect this PR and attempt to merge it before merging others (and releasing a beta.1)
As I see it, the problem is deeper: `cargo readme` and `cargo rdme` both have limitations/issues with intradoc links, so many links will fail (now and in the future). I can think of several solutions:

1.  in the past, I used a scripting solution (bash+awk) to do some links substitution, this is similar to your script. This almost works, but is clearly hackish and sometimes breaks
2. Remove links in the README part: that would be sad, since it looks like most crates do that :/
3. Wait for `cargo-rdme` to support intradoc links: this seems complex and not likely to happen soon
4. Use other tools based on `rustdoc` (unlike the above ones) like [cargo doc2readme](https://codeberg.org/msrd0/cargo-doc2readme)
    - pros: `doc2readme` seems to work here, at least I could generate a `README.md` file with all links working (result uploaded [here](https://gist.github.com/chifflier/e3832e60963b4acbb080fd7d824f75ab), see below for important note). It also has support for CI checks, and runs easily and automatically.
    - cons: it is not widely used, so there is always the risk to have a new breakage in some time
    - NOTE: the doc links use the version referenced in the crate (0.19.1-beta.1), which is not yet released. Thus, they are broken until this is released.

I would like to go forward and release a new version ASAP, so I think some compromise is necessary. What do you think of switching to doc2readme?

cpu: Thanks for doing this analysis.

> What do you think of switching to doc2readme?

It sounds like an OK option given the various trade-offs. :thumbsup: For what it's worth I've also had positive exchanges with the maintainer in other repos. 

> NOTE: the doc links use the version referenced in the crate (0.19.1-beta.1), which is not yet released. Thus, they are broken until this is released.

I also notice it doesn't seem to directly link to the items, but instead populates a search query. That's not ideal, but also doesn't seem like a deal breaker. Maybe there's a way to improve on that in the future.

> I would like to go forward and release a new version ASAP

Beyond the README issue I strongly think it makes sense to consider increasing MSRV with this release. There are many important crates with an MSRV of 1.85+ these days (reqwest, tokio, rand, quinn, rustls, hickory-dns, etc) and setting it to 1.85 would still cover what ships with Debian stable, while also addressing issues coming up in PRs (e.g. https://github.com/rusticata/x509-parser/pull/233). I personally don't think this crate is in a position where a more aggressive policy is helpful.

There are also a number of bug fix PRs I've approved that would be nice to include in a release soon. I've been hesitant to merge anything without your review, no matter how trivial.

chifflier: Thank you for your quick reply

> > What do you think of switching to doc2readme?
> 
> It sounds like an OK option given the various trade-offs. 👍 For what it's worth I've also had positive exchanges with the maintainer in other repos.

Good to know!
> 
> > NOTE: the doc links use the version referenced in the crate (0.19.1-beta.1), which is not yet released. Thus, they are broken until this is released.
> 
> I also notice it doesn't seem to directly link to the items, but instead populates a search query. That's not ideal, but also doesn't seem like a deal breaker. Maybe there's a way to improve on that in the future.

I will continue investigating a bit - it seems the tool is supposed to fall back on use search only on fall back cases, so maybe this can be improved in my example


> 
> > I would like to go forward and release a new version ASAP
> 
> Beyond the README issue I strongly think it makes sense to consider increasing MSRV with this release. There are many important crates with an MSRV of 1.85+ these days (reqwest, tokio, rand, quinn, rustls, hickory-dns, etc) and setting it to 1.85 would still cover what ships with Debian stable, while also addressing issues coming up in PRs (e.g. #233). I personally don't think this crate is in a position where a more aggressive policy is helpful.

The MSRV policy (now clarified in this PR :) ) states 12 months minimum, so 1.85 is acceptable.

> 
> There are also a number of bug fix PRs I've approved that would be nice to include in a release soon. I've been hesitant to merge anything without your review, no matter how trivial.

This is exactly the goal: I would like to first get back `master` in shape (fixing CI etc.), then sort issues/PRs to merge some of them before beta.1, and continue working. This is just the first step

cpu: Wonderful! I should be available this week if I can do anything to help.

chifflier: Ok, this starts to look like a never-ending story with too many new developments
It seems a new version of `cargo-rdme` was released [2.0.0](https://github.com/orium/cargo-rdme/releases/tag/v2.0.0) with support for extracting links using cargo-doc.
I did a quick test and it seems to resolve the problem, so I reverted my (local only, not pushed) experiment with `doc2readme` for now.
The good thing is that there is almost nothing to do with the existing PR: updating `cargo-rdme` is enough. I will push a commit with updated `README.md`, and this PR can start again being reviewed for merge.
Thanks

chifflier:
Ok, so, I do not really like that but I had to include a few other changes in the PR to fix urgent problems and attempt to have a cleaner CI
There is still an issue with `cargo-rdme` and nightly version. Geez, this is annoying

chifflier:
Ok, so, `cargo-rdme` pushed a fix/workaround, so everything should be fine now! CI is green :tada: 
@cpu : would you have some time to review this? This PR is blocking several others and it is becoming urgent
Thanks in advance

src/validate/extensions.rs
```diff
@@ -41,6 +41,8 @@ pub struct X509ExtensionsValidator;
 impl<'a> Validator<'a> for X509ExtensionsValidator {
     type Item = &'a [X509Extension<'a>];
 
+    // Tell clippy to _not_ suggest merging a long `if` check into an already long pattern match
+    #[allow(clippy::collapsible_match)]
```
cpu: Since the MSRV is > 1.81 now, I would recommend using `#[expect(clippy::collapsible_match)]` here so that it will warn in the future if this code is ever refactored so that the lint doesn't fire anymore (signalling that the attribute can be cleaned up).

src/validate/structure.rs
```diff
@@ -65,6 +65,8 @@ pub struct TbsCertificateStructureValidator;
 impl<'a> Validator<'a> for TbsCertificateStructureValidator {
     type Item = TbsCertificate<'a>;
 
+    // Tell clippy to _not_ suggest merging a long `if` check into an already long pattern match
+    #[allow(clippy::collapsible_match)]
```
cpu: ditto here: RE `allow` -> `expect`.

.github/workflows/rust.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       matrix:
         rust:
           - stable
-          - 1.85
+          - 1.88
```
cpu, line 18: I know I've argued for a higher MSRV in the past, but I was thinking ~1.85. That matches Rustls + Reqwest. I think Tokio is still on 1.71 or so.

I don't know that `time` alone is a sufficient motivator to take 1.88. AIUI [RUSTSEC-2026-0009](https://rustsec.org/advisories/RUSTSEC-2026-0009) is pretty narrow in its affected functions and I don't think x509-parser uses any of them. I'd be in favor of keeping MSRV at 1.85, pinning the `time` version with `Cargo.lock` to the last update with compatible MSRV, and encouraging downstream users to ignore RUSTSEC-2026-0009 for x509-parser as not applicable.

WDYT?

cpu (approved):
I haven't done an extensive review but I think given how long this has been blocked it makes sense to merge + iterate if anything new pops up.

src/validate/extensions.rs
chifflier: Oh, I did not know of `expect` for clippy. Added in next commit, thank you

src/validate/structure.rs
chifflier: Also changed

.github/workflows/rust.yml
chifflier: `time` has always been a difficult dependency to handle, because
- they tend to use very recent MSRVs in new releases, and they push security/bug fixes only in new releases :disappointed:
- using a locked older version is possible, but then `cargo audit` complains. Some downstream users mentioned their CI tools failing and requiring manual override

1.85 or 1.88 are both fine to me. I fear the next version of `time` will break things anyway.
I will have a look at 1.85

chifflier:
Thank you for the quick review!
I will now merge changes (including a revert to MSRV 1.85 as discussed) and close this
Let's use other issues/PRs if there is anything to discuss

cpu:
Great! Thanks :-)

PR #231, opened 2026-01-30
feat: add wasm-bindgen feature for wasm

Its-Just-Nans:
This MR add the wasm-bindgen feature of the `time` crate on wasm https://docs.rs/crate/time/latest/features 

Without the feature we have the error

```
time not implemented on this platform
```

Test file

```rust
//! cargo add --dev wasm-bindgen-test
//! wasm-pack test --node
#![cfg(target_arch = "wasm32")]

use x509_parser::pem::{parse_x509_pem, Pem};
use x509_parser::{parse_x509_certificate, x509::X509Version};

static IGCA_PEM: &[u8] = include_bytes!("../assets/IGC_A.pem");



#[wasm_bindgen_test::wasm_bindgen_test]
fn test_x509_parse_pem() {
    let pem: Vec<Pem> = Pem::iter_from_buffer(IGCA_PEM)
        .collect::<Result<_, _>>()
        .ok()
        .unwrap();
    let x509 = pem[0].parse_x509().unwrap();
    assert_eq!(x509.validity.is_valid(), false);
}
```

```sh
cargo install wasm-pack

cargo add --dev wasm-bindgen-test
wasm-pack test --node
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,6 +4,7 @@ on:
   push:
   pull_request:
   merge_group:
+  workflow_dispatch:
   schedule:
     - cron: '0 18 * * *'
 
@@ -118,3 +119,16 @@ jobs:
         uses: actions/checkout@v4
       - name: Check semver
         uses: obi1kenobi/cargo-semver-checks-action@v2
+
+  wasm-test:
+    name: Wasm test Suite
+    needs: check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - run: rustup target add wasm32-unknown-unknown
+      # TODO: If the MSRV is updated, we can remove this step and just add the dev-dependency to Cargo.toml
+      - run: cargo add wasm-bindgen-test --dev # ensure dependency is present for wasm tests
+      - run: cargo install wasm-pack --locked
+      - run: wasm-pack test --node
--- a/tests/test01.rs
+++ b/tests/test01.rs
@@ -1,17 +1,21 @@
-use nom::bytes::complete::take;
-
 #[test]
 fn test01() {
     let data = b"0\x88\xff\xff\xff\xff\xff\xff\xff\xff00\x0f\x02\x000\x00\x00\x00\x00\x00\x0000\x0f\x00\xff\x0a\xbb\xff";
     let _ = x509_parser::parse_x509_certificate(data);
 }
 
+#[cfg(not(target_arch = "wasm32"))]
 fn parser02(input: &[u8]) -> nom::IResult<&[u8], ()> {
+    use nom::bytes::complete::take;
     let (_hdr, input) = take(1_usize)(input)?;
     let (_data, input) = take(18_446_744_073_709_551_615_usize)(input)?;
     Ok((input, ()))
 }
 
+/// This test is not running on wasm32 because `parser02` needs a big usize
+/// Therefore on wasm32 target it triggers a compilation error:
+/// literal out of range for `usize``
+#[cfg(not(target_arch = "wasm32"))]
 #[test]
 fn test02() {
     let data = b"0\x88\xff\xff\xff\xff\xff\xff\xff\xff00\x0f\x02\x000\x00\x00\x00\x00\x00\x0000\x0f\x00\xff\x0a\xbb\xff";
--- a/tests/wasm32.rs
+++ b/tests/wasm32.rs
@@ -1,0 +1,26 @@
+//! This file is used to test the time crate with wasm.
+//! This needs the `wasm-bindgen-test` crate to compile
+//!
+//! ```sh
+//! # cargo add wasm-bindgen-test --dev
+//! wasm-pack test --node
+//! ```
+//!
+
+#![cfg(target_arch = "wasm32")]
+
+use x509_parser::pem::Pem;
+
+static IGCA_PEM: &[u8] = include_bytes!("../assets/IGC_A.pem");
+
+/// This test is used to check if the time crate is working correctly on wasm32 target
+/// because it is used in the x509 parser to check the validity of the certificate
+#[wasm_bindgen_test::wasm_bindgen_test]
+fn test_x509_parse_pem() {
+    let pem: Vec<Pem> = Pem::iter_from_buffer(IGCA_PEM)
+        .collect::<Result<_, _>>()
+        .ok()
+        .unwrap();
+    let x509 = pem[0].parse_x509().unwrap();
+    assert_eq!(x509.validity.is_valid(), false);
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,6 +125,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
+version = "3.19.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5dd9dc738b7a8311c7ade152424974d8115f2cdad61e8dab8dac9f2362298510"
+
+[[package]]
 name = "cc"
 version = "1.2.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -297,6 +303,16 @@ checksum = "38f262f097c174adebe41eb73d66ae9c06b2844fb0da69969647bbddd9b0538a"
 dependencies = [
  "getrandom 0.3.3",
  "libc",
+]
+
+[[package]]
+name = "js-sys"
+version = "0.3.85"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8c942ebf8e95485ca0d52d97da7c5a2c387d0e7f0ba4c35e93bfcaee045955b3"
+dependencies = [
+ "once_cell",
+ "wasm-bindgen",
 ]
 
 [[package]]
@@ -537,6 +553,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustversion"
+version = "1.0.22"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
+
+[[package]]
 name = "serde"
 version = "1.0.219"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -618,6 +640,7 @@ checksum = "8a7619e19bc266e0f9c5e6686659d394bc57973859340060a69221e57dbc0c40"
 dependencies = [
  "deranged",
  "itoa",
+ "js-sys",
  "num-conv",
  "powerfmt",
  "serde",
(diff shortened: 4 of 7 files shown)


Its-Just-Nans:
note that the `tests/test01.rs` needs to be restricted for non-wasm because usize is not the same in wasm

I can add that on this MR


Edit: DONE

cpu (dismissed):
Seems reasonable, but is there a way to add CI coverage so that support doesn't accidentally regress?

Its-Just-Nans:
> is there a way to add CI coverage so that support doesn't accidentally regress?

yes, I will add that

Its-Just-Nans:
Unfortunalty it's seems not possible because of the current MSRV

Or i can add dymanicly in the CI the 

```
cargo add --dev wasm-bindgen-test
```

tests/test01.rs
```diff
@@ -1,3 +1,8 @@
+//! This test is not running on wasm32 because `parser02` need a big usize
```
cpu: You have a small typo here: "because `parser02` need a" -> "because `parser02` needs a` (plural).

tests/test01.rs
```diff
@@ -1,3 +1,8 @@
...
+//! Therefore on wasm32 target it triggers a compilation error:
+//! literal out of range for `usize``
+#![cfg(not(target_arch = "wasm32"))]
```
cpu: If it's just `parser02` that's failing on wasm32, I think it would be an improvement to gate just that test instead of the whole file.

src/wasm32.rs
```diff
@@ -0,0 +1,25 @@
+//! this file is used to test the time crate with wasm
```
cpu: I'm not familiar with the wasm ecosystem so I might need some help here to understand why this is added as a `src/` module without any `test` gating, instead of a `tests/` integration test, or an example.

.github/workflows/rust.yml
```diff
@@ -5,7 +5,7 @@ on:
   pull_request:
   merge_group:
   schedule:
-    - cron: '0 18 * * *'
+    - cron: "0 18 * * *"
```
cpu: please avoid unrelated formatting changes inside a commit like this.

.github/workflows/rust.yml
```diff
@@ -118,3 +118,18 @@ jobs:
...
+  wasm-test:
+    name: Wasm test Suite
+    needs: check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/cache@v5
+        with:
+          path: target/
+          key: ${{ runner.os }}-${{ hashFiles('Cargo.lock') }}
```
cpu: I don't think any other CI tasks are using caching. I think it's a good idea, but would prefer to see that pursued separately and in a more uniform way.

tests/wasm32.rs
```diff
@@ -1,7 +1,6 @@
-//! this file is used to test the time crate with wasm
+//! This file is used to test the time crate with wasm
```
cpu: The changes in this file (particularly the `assert_eq` change) should be done in the commit that introduced the test.

.github/workflows/rust.yml
```diff
@@ -132,5 +132,6 @@ jobs:
           key: ${{ runner.os }}-${{ hashFiles('Cargo.lock') }}
       - uses: dtolnay/rust-toolchain@stable
       - run: rustup target add wasm32-unknown-unknown
+      - run: cargo add wasm-bindgen-test --dev --locked # ensure dependency is present
```
cpu: Why did you decide to remove the `Cargo.toml` dev-dependency in favour of this approach? It seems fine, but I was hoping to understand better (and perhaps it could be rolled into the earlier commit that adds CI).

.github/workflows/rust.yml
```diff
@@ -132,6 +132,6 @@ jobs:
           key: ${{ runner.os }}-${{ hashFiles('Cargo.lock') }}
       - uses: dtolnay/rust-toolchain@stable
       - run: rustup target add wasm32-unknown-unknown
-      - run: cargo add wasm-bindgen-test --dev --locked # ensure dependency is present
+      - run: cargo add wasm-bindgen-test --dev # ensure dependency is present for wasm tests
```
cpu, line 132: There's a small typo in your commit message, "no lcoked" -> "no locked". 

I'd also be interested in knowing why you want to avoid `--locked` here. Did it break something?

cpu (commented):
Thanks for iterating on this, I think it's getting close :-)

I don't know how others feel, but I'd personally have an easier time reviewing this if you were able to rebase to squash your own iteration down into the original commits.

.github/workflows/rust.yml
Its-Just-Nans: The CI failed

https://github.com/rusticata/x509-parser/actions/runs/21548752840/job/62094247918

```
Run cargo add wasm-bindgen-test --dev --locked
    Updating crates.io index
warning: ignoring wasm-bindgen-test@0.3.58 (which requires rustc 1.71) to maintain x509-parser's rust-version of 1.67.1
      Adding wasm-bindgen-test v0.3.55 to dev-dependencies
             Features:
             + std
             - gg-alloc
error: the manifest file /home/runner/work/x509-parser/x509-parser/Cargo.toml needs to be updated but --locked was passed to prevent this
Error: Process completed with exit code 101.
```

.github/workflows/rust.yml
Its-Just-Nans: I honestly don't like this solution. But the with MRSV current policy I can't add wasm-bindgen test. If you have an idea to solve that - that would be great !

https://github.com/rusticata/x509-parser/actions/runs/21535786463/job/62061103798
https://github.com/rusticata/x509-parser/pull/231#issuecomment-3827117195

```
error: package `wasm-bindgen-macro v0.2.108` cannot be built because it requires rustc 1.71 or newer, while the currently active rustc version is 1.67.1
Either upgrade to rustc 1.71 or newer, or use
cargo update -p wasm-bindgen-macro@0.2.108 --precise ver
where `ver` is the latest version of `wasm-bindgen-macro` supporting rustc 1.67.1
```

.github/workflows/rust.yml
cpu: Thanks, we're discussing similar MSRV issues [in this issue](https://github.com/rusticata/x509-parser/pull/233#issuecomment-3885988451). Dropping `--locked` for now makes sense :+1:

tests/test01.rs
Its-Just-Nans: No, the entire file is not compilable with wasm32 due to the usize being smaller in wasm 

`18_446_744_073_709_551_615_usize`

src/wasm32.rs
Its-Just-Nans: That's an old commit !

.github/workflows/rust.yml
cpu: For now this is OK, longer term I think we should try and convince the main maintainer that the MSRV should be increased to 1.71+.

Perhaps worth leaving a comment alongside this `cargo add` describing how we could switch to a dev-dep once the MSRV allows it so we don't forget.

.github/workflows/rust.yml
Its-Just-Nans: Old commit !

.github/workflows/rust.yml
Its-Just-Nans: I'm removing the cache

```
      - uses: actions/cache@v5
        with:
          path: target/
          key: ${{ runner.os }}-${{ hashFiles('Cargo.lock') }}
```

Its-Just-Nans:
**Thanks for the review!**

> I don't know how others feel, but I'd personally have an easier time reviewing this if you were able to rebase to squash your own iteration down into the original commits.

Okay, I understand that the commits history is missleading since I did multiple changes.
I cleaned the commits

btw a save of the commit order is available on https://github.com/Its-Just-Nans/x509-parser/tree/time-wasm-bindgen-save

tests/test01.rs
cpu: I'm not sure I see why `test01` is affected by that since it doesn't use the function that contains that usize. If both `parser02` and `test02` were conditionally compiled out for non-WASM, what would the issue with `test01` be?

(I don't want to get stuck on this since the test file itself seems a little dubious/under-documented. I'm happy to leave it the way you have it)

src/wasm32.rs
cpu: > That's an old commit !

I'd prefer to review the branch with a clean history with stand-alone commits vs having an implementation in one "old" commit, and then fixes spread out in separate follow-up commits.

Its-Just-Nans:
In my latest change

- conditional compilation on `parser02` and `test02` only https://github.com/rusticata/x509-parser/pull/231#discussion_r2809620746
- added a comment on the rust.yml for the dependency https://github.com/rusticata/x509-parser/pull/231#discussion_r2809593269

tests/test01.rs
Its-Just-Nans: you're right

cpu (approved):
Thanks!

I think you misunderstood me and squashed everything into one commit. That's better and I'm OK approving, but in the future my personal preference would be something like:

* a commit fixing the typos
* a commit gating the tests that won't run on wasm32
* a commit fixing the `time` dependency feature
* a commit adding the test & CI integration

Hope that makes sense! I know it's more work, but in my experience it optimizes for the reviewer and it's hard to find time for this crate already :-)

Its-Just-Nans:
Okay I understand, sorry for that

Should I bump the `time` dep since the security audit is failing ?

https://github.com/rusticata/x509-parser/actions/runs/22038536527/job/63675425805?pr=231

```

## Vulnerabilities

### [RUSTSEC-2026-0009](https://rustsec.org/advisories/RUSTSEC-2026-0009.html)

> Denial of Service via Stack Exhaustion

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `time`                      |
| Version             | `0.3.41`                   |
| URL                 | [https://github.com/time-rs/time/blob/main/CHANGELOG.md#0347-2026-02-05](https://github.com/time-rs/time/blob/main/CHANGELOG.md#0347-2026-02-05) |
| Date                | 2026-02-05                         |
| Patched versions    | `>=0.3.47`                  |
| Unaffected versions | `<0.3.6`               |

## Impact

When user-provided input is provided to any type that parses with the RFC 2822 format, a denial of
service attack via stack exhaustion is possible. The attack relies on formally deprecated and
rarely-used features that are part of the RFC 2822 format used in a malicious manner. Ordinary,
non-malicious input will never encounter this scenario.

## Patches

A limit to the depth of recursion was added in v0.3.47. From this version, an error will be returned
rather than exhausting the stack.

## Workarounds

Limiting the length of user input is the simplest way to avoid stack exhaustion, as the amount of
the stack consumed would be at most a factor of the length of the input.
```

cpu:
> Okay I understand, sorry for that

NP!

> Should I bump the time dep since the security audit is failing ?

No, it's OK to ignore that for now. There's a fix pending in https://github.com/rusticata/x509-parser/pull/223

PR #233, opened 2026-02-11
feat: support "verify" with FIPS crypto backend

domodwyer:
Hi there! First off, thanks for this library and the careful thought put into the security aspects.

I'm working on a system that is deployed in FedRAMP environments, which requires FIPS certified crypto backends, and fortunately the `aws-lc-rs` crate used by `x509-parser` is [already supported][fips]! However it's configured to use the non-FIPS backend when enabling the `verify-aws` feature today.

This PR adds a new feature flag `verify-aws-fips` to go alongside the existing `verify-aws` feature flag, enabling the FIPS-approved aws-lc-rs backend.

There's no code changes, just feature flag toggling :+1:

[fips]: https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/FIPS.md

---

* feat: support "verify" with FIPS crypto backend (8f355a0)
      
      Allow verifying cryptographic signatures using the AWS-LC's FIPS crypto
      backend.
      
      This change allows this library to be used in FedRAMP / US Gov
      deployments, which have a hard requirement on using FIPS-approved crypto
      modules only.
      
      The "verify-aws-fips" feature flag is functionally identical to using
      "verify-aws", but it selects the FIPS backend in aws-lc-rs.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -71,12 +71,24 @@ jobs:
           - --all-features
           - --features=verify
           - --features=verify-aws
+          - --features=verify-aws-fips
           - --features=validate
     steps:
       - uses: actions/checkout@v4
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable
       - run: cargo test --locked ${{ matrix.features }}
+
+  test_fips:
+    name: verify-aws-fips dependency on FIPS backend only
+    needs: check-all-features
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - run: "if cargo tree --features=verify-aws-fips -i aws-lc-sys; then false; else [ $? -eq 101 ]; fi"
+      - run: "cargo tree --features=verify-aws-fips -i aws-lc-fips-sys"
 
   fmt:
     name: Rustfmt
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
   merge_group:
   schedule:
-    - cron: '0 18 * * *'
+    - cron: "0 18 * * *"
 
 jobs:
   check:
@@ -87,7 +87,11 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable
+      # Ensure the non-FIPS aws-lc-sys crate is not part of the dependency tree.
+      #
+      # "cargo tree" exists 101 when queried for a crate that's not in the tree.
       - run: "if cargo tree --features=verify-aws-fips -i aws-lc-sys; then false; else [ $? -eq 101 ]; fi"
+      # Ensure the optional FIPS crypto module is depended.
       - run: "cargo tree --features=verify-aws-fips -i aws-lc-fips-sys"
 
   fmt:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -40,7 +40,6 @@ jobs:
     strategy:
       matrix:
         rust:
-          - stable
           - 1.85
           - nightly
     steps:
@@ -50,6 +49,9 @@ jobs:
           toolchain: ${{ matrix.rust }}
       - run: RUSTFLAGS="-D warnings" cargo check --locked --all-targets --all-features
 
+  # Also covers `check-all-features` on stable: the powerset includes every
+  # combination except pairing the two aws-lc-rs backends, and the
+  # `--all-features` pairing is exercised by the `test_features` job.
   feature-powerset:
     name: Check feature powerset
     runs-on: ubuntu-latest
@@ -59,16 +61,7 @@ jobs:
       - uses: taiki-e/install-action@cargo-hack
       # The two aws-lc-rs backends are never used together outside of
       # --all-features builds, so skip those combinations.
-      - run: RUSTFLAGS="-D warnings" cargo hack check --locked --feature-powerset --mutually-exclusive-features verify-aws,verify-aws-fips
-
-  test:
-    name: Test Suite
-    needs: check
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
-      - run: cargo test --locked --all-features
+      - run: RUSTFLAGS="-D warnings" cargo hack check --locked --all-targets --feature-powerset --mutually-exclusive-features verify-aws,verify-aws-fips
 
   test_features:
     name: Test suite (with features)
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -50,6 +50,17 @@ jobs:
           toolchain: ${{ matrix.rust }}
       - run: RUSTFLAGS="-D warnings" cargo check --locked --all-targets --all-features
 
+  feature-powerset:
+    name: Check feature powerset
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: taiki-e/install-action@cargo-hack
+      # The two aws-lc-rs backends are never used together outside of
+      # --all-features builds, so skip those combinations.
+      - run: RUSTFLAGS="-D warnings" cargo hack check --locked --feature-powerset --mutually-exclusive-features verify-aws,verify-aws-fips
+
   test:
     name: Test Suite
     needs: check
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
   merge_group:
   schedule:
-    - cron: '0 18 * * *'
+    - cron: "0 18 * * *"
 
 jobs:
   check:
@@ -87,7 +87,11 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable
+      # Ensure the non-FIPS aws-lc-sys crate is not part of the dependency tree.
+      #
+      # "cargo tree" exits 101 when queried for a crate that's not in the tree.
       - run: "if cargo tree --features=verify-aws-fips -i aws-lc-sys; then false; else [ $? -eq 101 ]; fi"
+      # Ensure the optional FIPS crypto module is depended.
       - run: "cargo tree --features=verify-aws-fips -i aws-lc-fips-sys"
 
   fmt:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,13 +38,14 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 default = []
-verify-aws = ["aws-lc-rs"]
+verify-aws = ["aws-lc-rs/aws-lc-sys"] # Non-FIPS backend
+verify-aws-fips = ["aws-lc-rs/fips"]  # FIPS crypto backend
 verify = ["ring"]
 validate = []
 
 [dependencies]
-aws-lc-rs = { version = "1.0", optional = true }
 asn1-rs = { version = "0.8.0-beta.1", features=["bigint", "datetime"] }
+aws-lc-rs = { version = "1.0", optional = true, default-features = false, features = ["alloc"] }
 data-encoding = "2.2.1"
 lazy_static = "1.4"
 nom = "8.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,13 +38,14 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 default = []
-verify-aws = ["aws-lc-rs"]
+verify-aws = ["aws-lc-rs/aws-lc-sys"] # Non-FIPS backend
+verify-aws-fips = ["aws-lc-rs/fips"]  # FIPS crypto backend
 verify = ["ring"]
 validate = []
 
 [dependencies]
-aws-lc-rs = { version = "1.0", optional = true }
 asn1-rs = { version = "0.8.0-beta.1", features=["bigint", "datetime"] }
+aws-lc-rs = { version = "1.0", optional = true, default-features = false, features = ["alloc"] }
 data-encoding = "2.2.1"
 nom = "8.0"
 oid-registry = { version="0.9.0-beta.1", features=["crypto", "x509", "x962"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,8 @@ include = [
 ]
 
 [package.metadata.docs.rs]
-all-features = true
+# all features except verify-aws-fips (aws-lc-fips-sys cannot build on docs.rs)
+features = ["validate", "verify", "verify-aws"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
--- a/src/certification_request.rs
+++ b/src/certification_request.rs
@@ -5,7 +5,11 @@ use crate::x509::{
     parse_signature_value, AlgorithmIdentifier, SubjectPublicKeyInfo, X509Name, X509Version,
 };
 
-#[cfg(any(feature = "verify", feature = "verify-aws"))]
+#[cfg(any(
+    feature = "verify",
+    feature = "verify-aws",
+    feature = "verify-aws-fips"
+))]
 use crate::verify::verify_signature;
 use asn1_rs::{
     BitString, DerParser, FromDer, Header, Input, Oid, OptTaggedImplicit, Sequence, Tag, Tagged,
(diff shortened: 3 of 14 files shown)


src/certification_request.rs
```diff
@@ -5,7 +5,11 @@ use crate::x509::{
     parse_signature_value, AlgorithmIdentifier, SubjectPublicKeyInfo, X509Name, X509Version,
 };
 
-#[cfg(any(feature = "verify", feature = "verify-aws"))]
+#[cfg(any(
+    feature = "verify",
+    feature = "verify-aws",
+    feature = "verify-aws-fips"
+))]
```
domodwyer, line 12: This is basically all the source code changes in this PR: adding a new feature flag in the same place as `verify-aws`.

Cargo.toml
```diff
@@ -38,13 +38,14 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 default = []
-verify-aws = ["aws-lc-rs"]
+verify-aws = ["aws-lc-rs/aws-lc-sys"] # Non-FIPS backend
+verify-aws-fips = ["aws-lc-rs/fips"]  # FIPS crypto backend
```
domodwyer, line 42: This is the new feature flag which switches on the `fips` feature of the `aws-lc-rs` crate.

It's either / or for the backends though, and the non-FIPS backend is enabled by default, so I've disabled the default features for aws-lc-rs and let the x509-parser feature flags enable the relevant crypto backend.

.github/workflows/rust.yml
```diff
@@ -71,13 +71,25 @@ jobs:
...
+    name: verify-aws-fips dependency on FIPS backend only
+    needs: check-all-features
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - run: "if cargo tree --features=verify-aws-fips -i aws-lc-sys; then false; else [ $? -eq 101 ]; fi"
+      - run: "cargo tree --features=verify-aws-fips -i aws-lc-fips-sys"
+
```
domodwyer, line 96: I figured it was a good idea to add a CI step to verify the FIPS backend is used when the fips flag is provided.

Specifically:

* The non-FIPS backend is **not** used (`aws-lc-sys`)
* The FIPS backend is used (`aws-lc-fips-sys`)

cpu:
Thanks for the PR (& self-review), I will try to review it soon but I'm supportive of the general direction.

> Security audit / security_audit (pull_request) - Failing after 3m

FWIW you can ignore this failing CI job, it's being addressed in https://github.com/rusticata/x509-parser/pull/223

domodwyer:
Awesome! No rush.

Thanks for the link - I see you're keeping support for a MSRV of 1.67.1 in the linked PR, but `aws-lc-rs` [wants at least 1.70](https://github.com/rusticata/x509-parser/actions/runs/21914235126/job/63276810151?pr=233):

> error: package `aws-lc-fips-sys v0.13.11` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.67.1

Should I gate the CI checks agains the FIPS backend with `if: matrix.rust != '1.67.1'`?

cpu:
> Thanks for the link - I see you're keeping support for a MSRV of 1.67.1 in the linked PR, but aws-lc-rs [wants at least 1.70](https://github.com/rusticata/x509-parser/actions/runs/21914235126/job/63276810151?pr=233)

I think this will ultimately come down to @chifflier's preference. IMHO I think an MSRV of 1.67 is too stale and many crates in the ecosystem with larger use have more aggressive MSRVs (e.g. Tokio at 1.71, aws-lc-rs at 1.70, rustls at 1.83). 

I would personally be in favour of taking an MSRV increase that was more in-line with those projects. 1.71 seems like an OK target and was a rust version released ~2 years ago.

domodwyer:
It's a shame there's no way of getting hold of rust version metadata for downloads from crates.io in order to make an informed decision - I agree it's quite a low MSRV and I doubt there's a meaningful fraction of users on <1.71 these days.

I will leave it as-is for now, and make any changes after review / when there's a MSRV plan :+1:

.github/workflows/rust.yml
```diff
@@ -71,13 +71,25 @@ jobs:
...
 
+  test_fips:
+    name: verify-aws-fips dependency on FIPS backend only
+    needs: check-all-features
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - run: "if cargo tree --features=verify-aws-fips -i aws-lc-sys; then false; else [ $? -eq 101 ]; fi"
```
cpu, line 93: Perhaps worth a comment here that exit status 101 is ["Cargo failed to complete"](https://doc.rust-lang.org/cargo/commands/cargo-tree.html#exit-status).

cpu (approved):
Thanks! Putting aside the question on how to handle MSRV this looks good to me otherwise.

cpu:
> 1.71 seems like an OK target and was a rust version released ~2 years ago.

Another point in favour: in would allow taking a dev dep on `wasm-bindgen` & associated bits for https://github.com/rusticata/x509-parser/pull/231

DarkmatterVale:
Hey all! Interested in seeing this deployed in the next release. Where does this stand?

cpu:
> Where does this stand?

This branch is targeting main, so unless backported to [the `x509-parser-0.18` branch](https://github.com/rusticata/x509-parser/tree/x509-parser-0.18) it would go out in a 0.19 beta, which as I understand things is blocked on https://github.com/rusticata/x509-parser/pull/223

domodwyer:
I'd be happy to backport to 0.18 if that helps? Would we have the same MSRV problem?

cpu:
> I'd be happy to backport to 0.18 if that helps?

It seems reasonable from my perspective, but I can't publish the crate so I think we're still blocked on chifflier (even setting aside the MSRV issue).

> Would we have the same MSRV problem?

Yes, I believe so.

.github/workflows/rust.yml
domodwyer: 48c57d1 :+1:

domodwyer:
> It's a shame there's no way of getting hold of rust version metadata for downloads from crates.io in order to make an informed decision - I agree it's quite a low MSRV and I doubt there's a meaningful fraction of users on <1.71 these days.

So fun fact, crates.io has a public Datadog dashboard that tracks requests by cargo version: https://p.datadoghq.com/sb/3a172e20-e9e1-11ed-80e3-da7ad0900002-973f4c1011257befa8598303217bfe3a?fromUser=false&refresh_mode=sliding&from_ts=1771511854139&to_ts=1772116654139&live=true

A quick check of this week of data shows that the highest ranking <1.71 version is 1.70 at 0.36%, and then after that, 1.65 at 0.07% of requests.

chifflier:
The MSRV has been upgraded to 1.85, so this part is now solved.
I'll see for merging the PR.
Thanks!

chifflier:
Hi,
I checkout out the PR and fixed conflicts introduced when merging other PRs. I am trying to build (`cargo build --all-features`) but get an error with `aws-lc-fips-sys`:
```
warning: aws-lc-fips-sys@0.13.16: Emitting configuration: cargo:rustc-cfg=x86_64_unknown_linux_gnu
warning: aws-lc-fips-sys@0.13.16: Only dynamic crypto library found in system install; using dynamic linking.
warning: aws-lc-fips-sys@0.13.16: Only dynamic crypto library found in system install; using dynamic linking.
warning: aws-lc-fips-sys@0.13.16: Building with: CMake
warning: aws-lc-fips-sys@0.13.16: Symbol Prefix: Some("aws_lc_fips_0_13_16")
warning: aws-lc-fips-sys@0.13.16: Target platform: 'x86_64-unknown-linux-gnu'
warning: aws-lc-fips-sys@0.13.16: Compilation of 'c11.c' succeeded - Ok(["/home/user/CODE/RUST/ASN1/x509-parser/target/debug/build/aws-lc-fips-sys-a6bbfe71212b012c/out/out-c11/7dfda64fdf5a526c-c11.o"]).
error: failed to run custom build command for `aws-lc-fips-sys v0.13.16`

... lots of lines ...

  In file included from /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/aws-lc-fips-sys-0.13.16/aws-lc/crypto/fipsmodule/bcm.c:41:
  /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/aws-lc-fips-sys-0.13.16/aws-lc/crypto/fipsmodule/../internal.h: In function 'OPENSSL_memchr':
  /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/aws-lc-fips-sys-0.13.16/aws-lc/crypto/fipsmodule/../internal.h:921:10: error: return discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
    921 |   return memchr(s, c, n);
        |          ^~~~~~
  At top level:
  cc1: note: unrecognized command-line option '-Wno-c11-extensions' may have been intended to silence earlier diagnostics
  cc1: all warnings being treated as errors
```

Any idea on how to fix this? Note: the C compiler used seems to be gcc 16, according to the logs

This also brings more problems:
- the build requires an executable TMPDIR (I could live with that, but only if this is required for _this_ feature, not in `--all-features` build (see next point)
- the `--all-features` becomes easily broken, and will add all dependencies in the result. Do someone have an idea on how to either override what is done in `Cargo.toml` for the `--all-features`? Otherwise, we may have to assume this and make all these features [mutually exclusive](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features). @cpu any thoughts?

Thanks

cpu:
> @cpu any thoughts?

I will take a look towards the end of this week.

cpu:
Hi, I pushed some revisions to this branch. It seemed easier than blocking on more back/forth given it has sat for a while. Sorry about that. Let me know if you'd prefer I don't push to your branch in the future.

> Any idea on how to fix this? Note: the C compiler used seems to be gcc 16

That failure is upstream, see https://github.com/aws/aws-lc-rs/issues/569

The FIPS module source is frozen for certification, so new-compiler fixes lag behind the non-FIPS `aws-lc-sys`. Until it's fixed upstream you can either build with `clang` (set `CC=clang CXX=clang++`) or use an older `gcc` (<= 13). GitHub's `ubuntu-latest` runners build it fine today using that latter strategy. FIPS is pain and nobody should subject themselves to it unless they have no choice :P

> the build requires an executable TMPDIR

This is also inherent to the `aws-lc-fips-sys` build, so on the upside only users that opt-in have to deal with it (and the heavy extra build deps, e.g. Go). Might be worth an upstream issue if it's something you want to work around.

> we may have to assume this and make all these features mutually exclusive

I'd recommend against. For `rustls`, `rcgen`, `hyper-rustls`, `tokio-rustls`, `instant-acme`, and `rustls-ffi` we model fips as an additive feature, and none use true mutually exclusive features / or a `compile_error!`.

Enabling `verify-aws` + `verify-aws-fips` together is how we manage `rustls`' fips configuration: both sys crates are compiled but `aws-lc-rs`'s own cfg selects the FIPS one, so `--all-features` stays correct, just heavyweight. Mutual exclusion would instead make `--all-features` fail to compile for every consumer. Ultimately feature flags kind of suck for this use-case but it's what we have.

> the --all-features becomes easily broken, and will add all dependencies in the result

I tried to address this a bit on the branch based on other projects:

- `docs.rs` metadata now lists explicit features instead of `all-features = true`, excluding `verify-aws-fips` (`docs.rs` can't run the Go/CMake build; rustls and instant-acme do the same, with the same comment). The feature still shows up in the doc-cfg annotations. It's a pain to keep this in sync, but I don't think there's a better way.
- New CI job: `cargo hack --feature-powerset --mutually-exclusive-features verify-aws,verify-aws-fips --all-targets`, with `-D warnings`. This is a "spiritual --all-features", it checks every feature combination except pairing the two `aws-lc-rs` backends. It already caught two real gate bugs (P-521 OID import and test fixture not gated to the aws features, also fixed on-branch).
- Redundant `--all-features` runs consolidated (dropped the duplicate test job and the stable row of check-all-features), cutting several `aws-lc-fips-sys` builds per CI run.
- ~~`verify-aws` now uses `aws-lc-rs`'s blessed `non-fips` alias (this is the same dependency graph, just more explicit).~~ _Oops: undid this, it actually enables the mutual-exclusivity I wanted to avoid_.

I also rebased on master and adjusted for the P-521 support that landed in the meantime (and fixed a typo in the CI config).

PTAL!

.github/workflows/rust.yml
```diff
@@ -50,14 +49,19 @@ jobs:
...
+    name: Check feature powerset
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
-      - run: cargo test --locked --all-features
+      - uses: taiki-e/install-action@cargo-hack
+      # The two aws-lc-rs backends are never used together outside of
+      # --all-features builds, so skip those combinations.
+      - run: RUSTFLAGS="-D warnings" cargo hack check --locked --all-targets --feature-powerset --mutually-exclusive-features verify-aws,verify-aws-fips
```
domodwyer, line 64: Ah this is a great idea!

.github/workflows/rust.yml
```diff
@@ -71,13 +75,29 @@ jobs:
...
+          - --features=verify-aws-fips
           - --features=validate
     steps:
       - uses: actions/checkout@v4
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable
       - run: cargo test --locked ${{ matrix.features }}
 
+  test_fips:
+    name: verify-aws-fips dependency on FIPS backend only
```
domodwyer, line 87: Thank you for preserving this property - it's helpful for my use case where we use cargo-deny to fail CI runs that add specific (non-FIPS) crates.

domodwyer (commented):
Looks great to me.

> Hi, I pushed some revisions to this branch. It seemed easier than blocking on more back/forth given it has sat for a while. Sorry about that. Let me know if you'd prefer I don't push to your branch in the future.

Absolutely fine by me - thank you for taking the time to get this into shape.

chifflier:
> Hi, I pushed some revisions to this branch. It seemed easier than blocking on more back/forth given it has sat for a while. Sorry about that. Let me know if you'd prefer I don't push to your branch in the future.

Hi @cpu,
Thanks a lot for your work! I tested locally, and everything looks good to me.

> The FIPS module source is frozen for certification, so new-compiler fixes lag behind the non-FIPS `aws-lc-sys`. Until it's fixed upstream you can either build with `clang` (set `CC=clang CXX=clang++`) or use an older `gcc` (<= 13). GitHub's `ubuntu-latest` runners build it fine today using that latter strategy. FIPS is pain and nobody should subject themselves to it unless they have no choice :P

Indeed, this is quite annoying. Let's just hope this does not crash the CI too often in the future

> > we may have to assume this and make all these features mutually exclusive
> 
> I'd recommend against. For `rustls`, `rcgen`, `hyper-rustls`, `tokio-rustls`, `instant-acme`, and `rustls-ffi` we model fips as an additive feature, and none use true mutually exclusive features / or a `compile_error!`.

Yes, I'd like to avoid making things exclusive. I think we'll keep the existing for now.
If this becomes too complex in the feature, we may switch to a different approach, like having a common `Trait` for cryptographic operations required to verify signatures and have multiple implementers depending on the features.

> > the --all-features becomes easily broken, and will add all dependencies in the result
> 
> I tried to address this a bit on the branch based on other projects:
> 
>     * `docs.rs` metadata now lists explicit features instead of `all-features = true`, excluding `verify-aws-fips` (`docs.rs` can't run the Go/CMake build; rustls and instant-acme do the same, with the same comment). The feature still shows up in the doc-cfg annotations. It's a pain to keep this in sync, but I don't think there's a better way.

I have no better idea, so it's fine

>     * New CI job: `cargo hack --feature-powerset --mutually-exclusive-features verify-aws,verify-aws-fips --all-targets`, with `-D warnings`. This is a "spiritual --all-features", it checks every feature combination except pairing the two `aws-lc-rs` backends. It already caught two real gate bugs (P-521 OID import and test fixture not gated to the aws features, also fixed on-branch).

Thanks

> 
>     * Redundant `--all-features` runs consolidated (dropped the duplicate test job and the stable row of check-all-features), cutting several `aws-lc-fips-sys` builds per CI run.

Fine

> 
>     * ~`verify-aws` now uses `aws-lc-rs`'s blessed `non-fips` alias (this is the same dependency graph, just more explicit).~ _Oops: undid this, it actually enables the mutual-exclusivity I wanted to avoid_.

Indeed, if we want to keep features additive, I think the best is to keep `aws-lc-rs` non-FIPS by default, and FIPS is the relevant feature is enabled.

> 
> 
> I also rebased on master and adjusted for the P-521 support that landed in the meantime (and fixed a typo in the CI config).

Thanks a lot!

chifflier:
Applied
Thank you @domodwyer  and @cpu

cpu:
> If this becomes too complex in the feature, we may switch to a different approach, like having a common Trait for cryptographic operations required to verify signatures and have multiple implementers depending on the features.

FWIW that's closer to what I was recommending in [this other thread](https://github.com/rusticata/x509-parser/pull/243#issuecomment-4084159883) about `RustCrypto`. I personally wouldn't be keen to add support for another cryptography library above/beyond ring and aws-lc-rs without doing that first.

PR #234, opened 2026-02-15
fix(time): use correct constructor for GeneralizedTime parsing

sumleo:
## Summary
- When parsing DER data with a `GeneralizedTime` tag, `from_der_content` was incorrectly calling `ASN1Time::new_utc()` instead of `ASN1Time::new_generalized()`
- This caused `is_generalizedtime()` to return `false` and `tag()` to return `Tag::UtcTime` for dates encoded as GeneralizedTime
- Resulted in false positive warnings during certificate structure validation (validator would warn that a GeneralizedTime-encoded date should use GeneralizedTime encoding)
- Added unit test to verify correct flag behavior after parsing GeneralizedTime content

## Test plan
- [x] New `test_generalizedtime_flags` test verifies correct `is_generalizedtime()`, `is_utctime()`, and `tag()` values


cpu (approved):

chifflier (approved):
Oh, seems I failed some copy/paste :)

Thank you for the PR

chifflier:
Applied, thanks

PR #235, opened 2026-02-15
fix(x509): set CONSTRUCTED=true for X509Name Tagged impl

sumleo:
## Summary
- `X509Name` is a SEQUENCE type, which is always constructed in ASN.1/DER encoding
- The `Tagged` trait implementation had `CONSTRUCTED = false`, which is incorrect
- This could cause tag matching issues when the framework validates or reconstructs tags for `X509Name`

## Test plan


cpu (approved):

chifflier:
Applied, thanks

PR #236, opened 2026-02-15
fix(public_key): correctly strip DER leading zero in RSA key_size()

sumleo:
## Summary
- `RSAPublicKey::key_size()` incorrectly calculated the key size by checking if the first byte's MSB was unset (`modulus[0] & 0x80 == 0`) and unconditionally stripping the first byte
- This produced incorrect results when the modulus had no DER leading zero padding (i.e., when the MSB of the actual modulus value happens to be 0)
- The fix checks specifically for a `0x00` padding byte before stripping it, and returns the key size for all non-empty modulus values (not just those with MSB unset)

## Test plan

### DIFF
--- a/src/public_key.rs
+++ b/src/public_key.rs
@@ -59,13 +59,17 @@ impl RSAPublicKey<'_> {
 
     /// Return the key size (in bits) or 0
     pub fn key_size(&self) -> usize {
-        if !self.modulus.is_empty() && self.modulus[0] & 0x80 == 0 {
-            // XXX len must substract leading zeroes
-            let modulus = &self.modulus[1..];
-            8 * modulus.len()
-        } else {
-            0
+        if self.modulus.is_empty() {
+            return 0;
         }
+        // DER encoding prepends a 0x00 byte to positive integers whose MSB is set.
+        // Strip only the leading zero padding byte, not arbitrary first bytes.
+        let modulus = if self.modulus[0] == 0x00 {
+            &self.modulus[1..]
+        } else {
+            self.modulus
+        };
+        8 * modulus.len()
     }
 }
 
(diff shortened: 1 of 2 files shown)


cpu (commented):
IMO this merits test coverage. I'd also prefer one commit instead of one with the change and one addressing a linting finding in the previous change.

src/public_key.rs
```diff
@@ -128,3 +132,58 @@ impl<'a> From<&'a [u8]> for ECPoint<'a> {
         ECPoint { data }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
cpu, line 137: I suppose this is better than nothing, but it seems conceivable you could generate real keys with the required properties and test them end-to-end through x509-parser as an integration test instead.

src/public_key.rs
```diff
@@ -59,13 +59,17 @@ impl RSAPublicKey<'_> {
 
     /// Return the key size (in bits) or 0
     pub fn key_size(&self) -> usize {
-        if !self.modulus.is_empty() && self.modulus[0] & 0x80 == 0 {
```
chifflier, line 62: It seems the previous code was written under the assumption a leading zero was always added (default case for `openssl`). Good catch!

src/public_key.rs
```diff
@@ -59,13 +59,17 @@ impl RSAPublicKey<'_> {
...
-            let modulus = &self.modulus[1..];
-            8 * modulus.len()
-        } else {
-            0
+        if self.modulus.is_empty() {
+            return 0;
         }
+        // DER encoding prepends a 0x00 byte to positive integers whose MSB is set.
+        // Strip only the leading zero padding byte, not arbitrary first bytes.
+        let modulus = if self.modulus[0] == 0x00 {
```
chifflier, line 67: Note that here we assume that the DER encoding is valid (i.e second byte cannot be zero, and length is > 1)
Since the parser is focused on accepting slightly incorrect constructs (to validate them later), I think this is acceptable.

src/public_key.rs
chifflier: Note: I did some tests but could not make `openssl` generate key modulus encoding without a leading zero.
Using `der2ascii + sed + ascii2der` I could craft a certificate and run tests, though.
I'll see if I can add a proper version of this in `assets/` later

chifflier (approved):
Looks good to me

chifflier:
Applied, thanks

PR #237, opened 2026-02-15
fix(certificate): add try_is_ca() to surface extension parsing errors

sumleo:
## Summary
- `is_ca()` used `unwrap_or(None)` which silently swallowed errors from `basic_constraints()`, such as `DuplicateExtensions`
- This could cause a certificate with malformed or duplicate basic constraints extensions to be silently treated as non-CA instead of surfacing the error
- Added `try_is_ca()` that propagates errors via `Result<bool, X509Error>`, allowing callers to distinguish between "not a CA" and "error reading extensions"
- `is_ca()` is preserved for backward compatibility, now delegating to `try_is_ca().unwrap_or(false)`

## Test plan
- [x] `is_ca()` behavior is unchanged for valid certificates (backward compatible)

### DIFF
--- a/src/certificate.rs
+++ b/src/certificate.rs
@@ -585,11 +585,19 @@ impl<'a> TbsCertificate<'a> {
     }
 
     /// Returns true if certificate has `basicConstraints CA:true`
+    ///
+    /// This is a convenience wrapper around [`try_is_ca`](Self::try_is_ca)
+    /// that returns `false` on error. Use `try_is_ca` if you need to
+    /// distinguish between "not a CA" and "error reading extensions".
     pub fn is_ca(&self) -> bool {
-        self.basic_constraints()
-            .unwrap_or(None)
-            .map(|ext| ext.value.ca)
-            .unwrap_or(false)
+        self.try_is_ca().unwrap_or(false)
+    }
+
+    /// Returns `Ok(true)` if certificate has `basicConstraints CA:true`,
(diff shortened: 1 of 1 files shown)


src/certificate.rs
```diff
@@ -585,11 +585,22 @@ impl<'a> TbsCertificate<'a> {
...
+    }
+
+    /// Returns `Ok(true)` if certificate has `basicConstraints CA:true`,
+    /// `Ok(false)` if the extension is absent or `CA` is false,
+    /// or an error if the extension could not be read (e.g. duplicate extensions).
+    pub fn try_is_ca(&self) -> Result<bool, X509Error> {
+        match self.basic_constraints()? {
+            Some(ext) => Ok(ext.value.ca),
+            None => Ok(false),
+        }
```
cpu: nit: I think this is clearer hoisting the `Ok()` and using the `map_or` combinator to stay closer to the style of the original impl:
 
```rust
pub fn try_is_ca(&self) -> Result<bool, X509Error> {
    Ok(self.basic_constraints()?.map_or(false, |ext| ext.value.ca))
}
```

cpu (commented):
This seems OK but in my opinion also deserves new test coverage. Try [der-ascii](https://github.com/google/der-ascii) if you can't do it easily programatically.

cpu (approved):

chifflier:
Applied, thanks

PR #238, opened 2026-02-15
fix(cri): iterate SET contents in iter_raw_values(), not SET TLV

sumleo:
## Summary
- `X509CriAttribute::iter_raw_values()` was iterating over `self.value` which contains the full SET TLV (tag + length + content)
- This caused the iterator to yield the entire SET as a single item rather than iterating over the individual `AttributeValue` entries inside it
- Fixed by parsing past the SET header before creating the iterator, so it correctly yields each value in the SET

## Test plan

### DIFF
--- a/src/cri_attributes.rs
+++ b/src/cri_attributes.rs
@@ -33,10 +33,21 @@ impl<'a> X509CriAttribute<'a> {
     }
 
     /// Iterate over the unparsed values of 'SET OF AttributeValue'
+    ///
+    /// Each item yields the raw DER content of an individual value within the SET,
+    /// returned as an [`Any`] object containing the DER-encoded tag, length, and value.
+    ///
+    /// If the SET header cannot be parsed, an empty iterator is returned.
     pub fn iter_raw_values(
         &self,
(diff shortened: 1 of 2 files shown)


src/cri_attributes.rs
```diff
@@ -36,7 +36,13 @@ impl<'a> X509CriAttribute<'a> {
     pub fn iter_raw_values(
         &self,
     ) -> impl Iterator<Item = Result<(Input<'a>, Any<'a>), BerError<Input<'a>>>> {
-        AnyIterator::<DerMode>::new(self.value.clone())
+        // `self.value` contains the full SET TLV; parse past the SET header
+        // to iterate over the individual values inside the SET
+        let content = match <Set>::parse_der_as_input(self.value.clone()) {
+            Ok((_, (_, content))) => content,
+            Err(_) => Input::default(),
```
cpu, line 48: This seems reasonable for this context, but might be worth mentioning in the rustdoc.

cpu (commented):
I think this fix is another candidate where test coverage would be helpful.

cpu (approved):

chifflier:
Applied, thanks

PR #239, opened 2026-02-15
fix(pem): clear line buffer after InvalidData and validate END label

sumleo:
## Summary
Two fixes in PEM parsing:

1. **Missing line buffer clear after InvalidData**: After encountering an `InvalidData` error (non-UTF-8 data in comment section), the `line` buffer was not cleared before continuing the loop. Since `read_line` appends to the buffer, subsequent reads would accumulate partial data from the failed read, potentially corrupting the `-----BEGIN` header detection.

2. **END label not validated against BEGIN label**: The END marker was accepted without verifying its label matches the BEGIN label. For example, a PEM block starting with `-----BEGIN CERTIFICATE-----` would accept `-----END PRIVATE KEY-----` as its terminator. Now validates that the END label matches.

## Test plan
- [x] Existing PEM tests continue to pass (valid PEM files have matching labels)

### DIFF
--- a/src/pem.rs
+++ b/src/pem.rs
@@ -123,6 +123,7 @@ impl Pem {
                 Ok(line) => line,
                 Err(e) if e.kind() == ErrorKind::InvalidData => {
                     // some tools put invalid UTF-8 data in PEM comment section. Ignore line
+                    line.clear();
                     continue;
                 }
                 Err(e) => {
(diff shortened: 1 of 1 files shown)


cpu:
Without weighing in on the fix or the diff yet this needs a reformat and I think should come with test coverage for both fixes. My preference would be one commit per fix + accompanying test.

chifflier (approved):

chifflier:
Applied, thanks

PR #240, opened 2026-02-15
fix(validate): return false from name validator on charset errors

sumleo:
## Summary
- `X509NameStructureValidator::validate()` always returned `true` regardless of validation results
- The `res` variable was declared as immutable `true` and never updated, even when invalid charset errors were detected and logged
- Fixed by making `res` mutable and setting it to `false` when non-ASCII characters are found in `PrintableString` or `IA5String` attribute values

## Test plan

### DIFF
--- a/src/validate/name.rs
+++ b/src/validate/name.rs
@@ -9,7 +9,7 @@ impl<'a> Validator<'a> for X509NameStructureValidator {
     type Item = X509Name<'a>;
 
     fn validate<L: Logger>(&self, item: &'a Self::Item, l: &'_ mut L) -> bool {
-        let res = true;
+        let mut res = true;
         // subject/issuer: verify charsets
         // - wildcards in PrintableString
         // - non-IA5 in IA5String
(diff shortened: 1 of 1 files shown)


cpu (commented):
I'm a little bit surprised that clippy didn't catch this 🤔 

This feels like a good candidate for a unit test to prevent regression.

src/validate/name.rs
```diff
@@ -30,3 +31,80 @@ impl<'a> Validator<'a> for X509NameStructureValidator {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::x509::{AttributeTypeAndValue, RelativeDistinguishedName, X509Name};
+    use asn1_rs::{Any, Input, Tag};
+
+    /// Helper to build an X509Name with a single attribute using the given tag and data.
+    fn make_name(tag: Tag, data: &[u8]) -> X509Name<'_> {
```
cpu, line 42: Was it not feasible to make real test certificates with the problematic data?

chifflier:
> I'm a little bit surprised that clippy didn't catch this 🤔
Indeed, it caused no error (since value was never changed), but maybe it could have been detected.

src/validate/name.rs
chifflier: Probably, though it would require a full certificate just to test for invalid charset in a specific location. It seems reasonable to have a specific unit test.
I'll see if I can find some certificates in the future.

chifflier (approved):

chifflier:
Applied, thanks

PR #241, opened 2026-02-15
fix: remove dead code, fix ns_comment remainder, and correct doc comments

sumleo:
## Summary
Three small fixes:

1. **Dead code in extension parser** (`extensions/mod.rs`): Removed `rem.take(rem.input_len())` which had no effect — `take` returns a new value without mutating `rem`, so the result was silently discarded

2. **Incorrect remainder in ns_comment fallback** (`extensions/ns_comment.rs`): The fallback parser for non-IA5String-wrapped comments used `input.take(input.input_len())` as the remainder, which returns the full input instead of an empty slice. Fixed to use `input.take_from(input.input_len())` consistent with the pattern used in `generalname.rs`

3. **Wrong doc comments** (`certification_request.rs`): Four doc comments on `X509CertificationRequestInfo` methods incorrectly referred to "CRL entry extensions" instead of "CSR attributes"

## Test plan

### DIFF
--- a/src/certification_request.rs
+++ b/src/certification_request.rs
@@ -192,26 +192,26 @@ pub struct X509CertificationRequestInfo<'a> {
 }
 
 impl X509CertificationRequestInfo<'_> {
-    /// Get the CRL entry extensions.
+    /// Get the CSR attributes.
     #[inline]
     pub fn attributes(&self) -> &[X509CriAttribute<'_>] {
         &self.attributes
     }
 
-    /// Returns an iterator over the CRL entry extensions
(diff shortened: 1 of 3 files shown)


cpu (commented):
These fixes seem OK to me, and I don't know that they rise to the point of requiring new unit test coverage.

I'd personally prefer to see this branch as three discrete commits, one per fix with an appropriate message, and no follow-on commits reformatting or updating imports.

cpu (approved):

chifflier:
Applied, thanks

PR #242, opened 2026-02-15
fix(validate): reject unsupported critical extensions per RFC 5280

sumleo:
## Summary
- The structure validator only warned about unsupported extensions without distinguishing between critical and non-critical ones
- Per RFC 5280 Section 4.2: "If a certificate contains a critical extension that is not recognized by the implementation, then the implementation MUST reject the certificate"
- Now reports unsupported **critical** extensions as errors (causing validation failure) while keeping non-critical unsupported extensions as warnings

## Test plan


cpu (commented):
The change makes sense to me, but I think this would be straight forward to add test coverage for.

cpu (approved):

chifflier:
Applied, thanks

PR #249, opened 2026-06-10
Reject CONTEXT-SPECIFIC tag class in serial number parser

### LINKED ISSUE #248: ASN.1 tag class not validated — UNIVERSAL vs CONTEXT-SPECIFIC confusion (X.690 §8.1.2)
The parser doesn't validate that ASN.1 tag classes match expectations. Specifically, when parsing X.509 structures, context-specific tags (class 10) in positions where UNIVERSAL tags (class 00) are expected are accepted without error.

X.690 §8.1.2 defines tag classes, and mixing them changes the semantic meaning of the element. Other implementations (OpenSSL, GnuTLS, wolfSSL, Botan, NSS) reject certificates with mismatched tag classes.

Found through 14-implementation differential fuzzing (PathDiff diff_fuzz_v2).

afldl:
## Problem

`parse_serial()` checks that the tag number equals `Tag::Integer` but does not verify the tag class. A CONTEXT-SPECIFIC `[2]` tag byte (`0x82`) has the same tag number as UNIVERSAL INTEGER (`0x02`) but a different class per X.690 §8.1.2. This makes the parser accept certificates with mismatched tag classes.

13 of 14 tested X.509 implementations reject this mutation. Only x509-parser and mbedTLS accept. See #248.

## Fix

Check `any.class() == Class::Universal` before comparing the tag number.

```diff
     let (rem, any) = Any::parse_der(input).map_err(|_| X509Error::InvalidSerial)?;
+    if any.class() != Class::Universal {
+        return Err(Err::Error(X509Error::InvalidSerial));
+    }
     any.tag()
         .assert_eq(Tag::Integer)
```

## Test

Single-byte PoC: offset 0x0d of a standard RSA-2048 self-signed cert, tag byte changed from `0x02` to `0x82`.

Test added: `test_serial_rejects_context_specific_tag`. All 66 existing tests pass.

### DIFF
--- a/src/validate/extensions.rs
+++ b/src/validate/extensions.rs
@@ -97,11 +97,10 @@ impl<'a> Validator<'a> for X509ExtensionsValidator {
                     test_critical!(SHOULD NOT ext, l, "SubjectAltName");
                     for name in san.general_names() {
                         match name {
-                            GeneralName::DNSName(ref s) | GeneralName::RFC822Name(ref s) => {
-                                // should be an ia5string
-                                if !s.as_bytes().iter().all(u8::is_ascii) {
-                                    l.warn(&format!("Invalid charset in 'SAN' entry '{s}'"));
-                                }
+                            GeneralName::DNSName(ref s) | GeneralName::RFC822Name(ref s)
+                                if !s.as_bytes().iter().all(u8::is_ascii) =>
+                            {
+                                l.warn(&format!("Invalid charset in 'SAN' entry '{s}'"));
                             }
                             _ => (),
                         }
--- a/src/validate/structure.rs
+++ b/src/validate/structure.rs
@@ -140,11 +140,10 @@ impl<'a> Validator<'a> for TbsCertificateStructureValidator {
             if let ParsedExtension::SubjectAlternativeName(san) = ext.parsed_extension() {
                 for name in san.general_names() {
                     match name {
-                        GeneralName::DNSName(ref s) | GeneralName::RFC822Name(ref s) => {
-                            // should be an ia5string
-                            if !s.as_bytes().iter().all(u8::is_ascii) {
-                                l.warn(&format!("Invalid charset in 'SAN' entry '{s}'"));
-                            }
+                        GeneralName::DNSName(ref s) | GeneralName::RFC822Name(ref s)
+                            if !s.as_bytes().iter().all(u8::is_ascii) =>
+                        {
+                            l.warn(&format!("Invalid charset in 'SAN' entry '{s}'"));
                         }
                         _ => (),
                     }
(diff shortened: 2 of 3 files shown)


cpu:
You can ignore the clippy & audit failings in CI. I believe both are going to be fixed by https://github.com/rusticata/x509-parser/pull/223 Please just keep the commits related to the issue/fix.

src/validate/extensions.rs
```diff
@@ -97,11 +97,10 @@ impl<'a> Validator<'a> for X509ExtensionsValidator {
...
                         match name {
-                            GeneralName::DNSName(ref s) | GeneralName::RFC822Name(ref s) => {
-                                // should be an ia5string
-                                if !s.as_bytes().iter().all(u8::is_ascii) {
-                                    l.warn(&format!("Invalid charset in 'SAN' entry '{s}'"));
-                                }
+                            GeneralName::DNSName(ref s) | GeneralName::RFC822Name(ref s)
+                                if !s.as_bytes().iter().all(u8::is_ascii) =>
+                            {
+                                l.warn(&format!("Invalid charset in 'SAN' entry '{s}'"));
```
chifflier, line 103: This change seems unrelated to the problem (and `cargo fmt` does not give the same result, FYI). Please remove this from the patch

src/validate/structure.rs
```diff
@@ -140,11 +140,10 @@ impl<'a> Validator<'a> for TbsCertificateStructureValidator {
...
                     match name {
-                        GeneralName::DNSName(ref s) | GeneralName::RFC822Name(ref s) => {
-                            // should be an ia5string
-                            if !s.as_bytes().iter().all(u8::is_ascii) {
-                                l.warn(&format!("Invalid charset in 'SAN' entry '{s}'"));
-                            }
+                        GeneralName::DNSName(ref s) | GeneralName::RFC822Name(ref s)
+                            if !s.as_bytes().iter().all(u8::is_ascii) =>
+                        {
+                            l.warn(&format!("Invalid charset in 'SAN' entry '{s}'"));
```
chifflier, line 146: Ditto (unrelated)

src/x509.rs
```diff
@@ -612,6 +612,10 @@ pub(crate) fn parse_serial(input: Input<'_>) -> IResult<Input<'_>, (&[u8], BigUi
     // RFC 5280 4.1.2.2: "The serial number MUST be a positive integer"
     // however, many CAs do not respect this and send integers with MSB set,
     // so we do not use `as_biguint()`
+    // X.690 §8.1.2: tag class must be UNIVERSAL for INTEGER
+    if any.class() != Class::Universal {
+        return Err(Err::Error(X509Error::InvalidSerial));
+    }
```
chifflier, line 618: To stay consistent with the next test, this should be changed to use `.class().assert_eq()` (see suggestion)

```suggestion
    any.class()
        .assert_eq(Class::Universal)
        .map_err(|_| X509Error::InvalidSerial)?;
```

chifflier (changes requested):
Hi,
Thank you for the report and PR.
The parser is indeed a bit permissive, and the proposed fix is correct.
I just suggested a few changes - can you update the PR?

Issue #251, opened 2026-07-19
impl Add<Duration> for ASN1Time panics on date overflow despite an Option return type


scadastrangelove:
# ISSUE — x509-parser: `ASN1Time + Duration` panics on overflow instead of returning `None`

**Title:** `impl Add<Duration> for ASN1Time` panics on date overflow despite an `Option` return type

### Bug

In `src/time.rs`, `impl Add<Duration> for ASN1Time` has return type `Option<ASN1Time>` but cannot ever
return `None` — instead it panics on date overflow:

```rust
fn add(self, rhs: Duration) -> Option<ASN1Time> {
    Some(ASN1Time::new(self.time + rhs))   // OffsetDateTime + Duration panics on overflow
}
```

`OffsetDateTime + Duration` is `checked_add(...).expect(...)` internally, so `add` aborts the thread
where its signature promises a `None`. The sibling `impl Sub<ASN1Time>` right below already guards
correctly and returns `None`.

### Reachability / severity

`self.time` can be attacker-influenced: a certificate `notAfter` of `99991231235959Z` (RFC 5280
§4.1.2.5's conventional "no well-defined expiration") parses to a max-year `Date`. x509-parser itself
never calls `Add` (its own `time_to_expiration` uses the checked `Sub`), so a live crash requires
downstream code doing something like `cert.validity().not_after + grace_period` on an untrusted cert —
a natural "is this cert valid N days from now?" pattern. I'd frame this as a **correctness/robustness**
bug with a latent DoS in consumer code, not an exploitable vulnerability in x509-parser itself. Not
requesting a CVE/RustSec advisory.

### Repro (against current default features)

```rust
use x509_parser::time::ASN1Time;
use time::{Duration, macros::datetime};

let t = ASN1Time::from(datetime!(9999-12-31 23:59:59 UTC));
let _ = t + Duration::days(1);   // panics: "resulting value is out of range"
```

### Fix (verified: compiles; max-year+1d → `None`; normal case unchanged)

```diff
     fn add(self, rhs: Duration) -> Option<ASN1Time> {
-        Some(ASN1Time::new(self.time + rhs))
+        self.time.checked_add(rhs).map(ASN1Time::new)
     }
```

PR incoming with this fix + a regression test.

### Minor secondary note (info only, no fix requested)

Duplicate CRL extensions are accepted by the parser (`many0`, no dedup), so
`CertificateRevocationList::crl_number()` first-matches while `extensions_map()` returns
[...]

PR #252, opened 2026-07-19
fix(time): return None on ASN1Time + Duration overflow instead of panicking

scadastrangelove:
# fix(time): return `None` on `ASN1Time + Duration` overflow instead of panicking

## What

`impl Add<Duration> for ASN1Time` (`src/time.rs`) has return type `Option<ASN1Time>`, but the
body was `Some(ASN1Time::new(self.time + rhs))`. `OffsetDateTime + Duration` panics on overflow
(internally `checked_add(...).expect(...)`), so `add` aborts the thread in exactly the case its
`Option` signature exists to report. The sibling `impl Sub<ASN1Time>` right below already guards
and returns `None`.

```diff
     fn add(self, rhs: Duration) -> Option<ASN1Time> {
-        Some(ASN1Time::new(self.time + rhs))
+        self.time.checked_add(rhs).map(ASN1Time::new)
     }
```

## Why it matters

`self.time` can come from a parsed certificate — e.g. a `notAfter` of `99991231235959Z`
(RFC 5280 §4.1.2.5's conventional "no well-defined expiration"), which parses to a max-year
date. Downstream code doing `cert.validity().not_after + grace_period` on an untrusted
certificate — a natural "is this cert valid N days from now?" pattern — then panics instead of
getting the `None` the API promises.

I'd frame this as a **correctness/robustness** fix rather than a security vulnerability:
`x509-parser` itself never calls `Add` (its `time_to_expiration` uses the checked `Sub`), so
the impact is confined to downstream callers that use the operator. Filed alongside issue #251.

## Test

Adds `time::tests::test_add_duration_overflow_returns_none`:

- `ASN1Time::from(datetime!(9999-12-31 23:59:59 UTC)) + Duration::days(1)` → `None` (was: panic);
- a normal date `+ Duration::days(1)` → `Some(..)` (unchanged).

Verified locally: `cargo test --lib time::` → `3 passed; 0 failed`.

Found & fixed with the [rust-in-peace](https://github.com/scadastrangelove/rust-in-peace) security pipeline.

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -197,7 +197,7 @@ impl Add<Duration> for ASN1Time {
 
     #[inline]
     fn add(self, rhs: Duration) -> Option<ASN1Time> {
-        Some(ASN1Time::new(self.time + rhs))
+        self.time.checked_add(rhs).map(ASN1Time::new)
     }
 }
 
@@ -223,6 +223,7 @@ impl From<OffsetDateTime> for ASN1Time {
 #[cfg(test)]
 mod tests {
     use time::macros::datetime;
+    use time::Duration;
 
     use super::ASN1Time;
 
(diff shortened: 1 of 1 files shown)


chifflier (approved):

chifflier:
Applied, thanks!

PR #230, opened 2025-12-15
docs: fix small README typos

cpu:
See https://github.com/rusticata/x509-parser/pull/229#discussion_r2619796911

PR #253, opened 2026-07-22
Fix typo in README.md (references #230)

chifflier:
Followup from #230 for changes not merged in #223

Using gui to edit file in #230 was not a good idea.
Anyway, `README.md` changes go through `src/lib.rs`, so re-creating the patch


cpu (approved):
LGTM, thank you for picking this up.

PR #255, opened 2026-07-22
CI: use rust 1.85 for MSRV version (missed update in last MSRV change)

chifflier:
The rust version in CI was not updated correctly when switching back to 1.85


cpu (approved):
Oops, good catch 👍

PR #257, opened 2026-08-06
Fix/pre commit and clippy

chifflier:
QA+CI improvements:
- add pre-commit configuration (I use [`prek`](https://github.com/j178/prek) which is compatible but faster) to run hooks locally on commits
- check the same hooks in CI
- centralize lints (rust and clippy) in `Cargo.toml` and enable more warnings
- fix all warnings and/or add bypasses in tests and examples

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,6 +8,19 @@ on:
     - cron: "0 18 * * *"
 
 jobs:
+  pre_commit:
+    timeout-minutes: 5
+    name: prek
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      # tools used in pre-commit hooks
+      - id: cargo-rdme
+        uses: rusticata/ci-action-rdme@v1
+      # /tools
+      - id: prek
+        uses: rusticata/ci-action-prek@v1
+
   check:
     name: Check
     runs-on: ubuntu-latest
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,7 +16,9 @@ jobs:
       - uses: actions/checkout@v6
       # tools used in pre-commit hooks
       - id: cargo-rdme
-        uses: rusticata/ci-action-rdme@v1
+        uses: rusticata/ci-action-rdme@v2
+        with:
+          rdme-args: --intralinks-features validate,verify,verify-aws
       # /tools
       - id: prek
         uses: rusticata/ci-action-prek@v1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,18 @@ time = { version="0.3.41", features=["formatting"] }
 [dev-dependencies]
 hex-literal = "0.4"
 
+[lints.rust]
+unsafe_code = "forbid"
+#missing_docs = "warn"
+unstable_features = "deny"
+unused_import_braces = "deny"
+unused_qualifications = "deny"
+missing_debug_implementations = "warn"
+unreachable_pub = "warn"
+
+[lints.rustdoc]
+broken_intra_doc_links = "deny"
+
 [lints.clippy]
 #pedantic = { level = "warn", priority = -1 }
 unwrap_used = "warn"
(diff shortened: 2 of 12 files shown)


.github/workflows/rust.yml
```diff
@@ -8,6 +8,19 @@ on:
     - cron: "0 18 * * *"
 
 jobs:
+  pre_commit:
```
cpu, line 11: I'm not sure I understand why it makes sense to run this in CI. AFAICT the pre-commit hook is duplicating checks that already exist in CI, so now we're doing them twice (?)

Cargo.toml
```diff
@@ -58,3 +58,14 @@ time = { version="0.3.41", features=["formatting"] }
 [dev-dependencies]
 hex-literal = "0.4"
 
+[lints.clippy]
+#pedantic = { level = "warn", priority = -1 }
```
cpu, line 62: Why land commented-out config? I think it'd be tidier to remove this and the other `print_stdout`/`print_stderr` entries if they're not being used.

Cargo.toml
```diff
@@ -58,6 +58,18 @@ time = { version="0.3.41", features=["formatting"] }
 [dev-dependencies]
 hex-literal = "0.4"
 
+[lints.rust]
+unsafe_code = "forbid"
+#missing_docs = "warn"
```
cpu, line 63: same feedback to consider here RE: not adding commented out config.

.github/workflows/rust.yml
```diff
@@ -16,7 +16,9 @@ jobs:
       - uses: actions/checkout@v6
       # tools used in pre-commit hooks
       - id: cargo-rdme
-        uses: rusticata/ci-action-rdme@v1
+        uses: rusticata/ci-action-rdme@v2
+        with:
+          rdme-args: --intralinks-features validate,verify,verify-aws
```
cpu, line 21: It'd make for a tidier commit history if this were folded into the commit that added the pre-commit config 👼

tests/readcsr.rs
```diff
@@ -192,14 +193,15 @@ fn read_csr_verify() {
...
 #[test]
 fn read_csr_with_custom_extension() {
-    let der = pem::parse_x509_pem(CSR_CUSTOM_EXTENSION).unwrap().1;
+    let der = parse_x509_pem(CSR_CUSTOM_EXTENSION).unwrap().1;
     let (rem, csr) = X509CertificationRequest::from_der(&der.contents)
         .expect("Could not parse CSR with custom extension");
 
     assert!(rem.is_empty());
-    dbg!(csr.certification_request_info.attributes());
+    //dbg!(csr.certification_request_info.attributes());
```
cpu, line 204: I would vote to remove this line instead of commenting it out.

Cargo.toml
chifflier: the objective is to un-comment them progressively. If we activate the warnings right now, there are a lot of fixes

.github/workflows/rust.yml
chifflier: It can be done, I just thought of these commits as slightly different (one for pre-commit and one for `rdme`) which could have been independant

tests/readcsr.rs
chifflier: agreed, I will update

Cargo.toml
cpu: I understand not wanting to activate it immediately, but I don't know that I see the value in leaving it commented out as some kind of reminder. Wouldn't a GitHub issue be better for tracking the work required to enable pedantic clippy if that's a goal?

.github/workflows/rust.yml
chifflier: I agree that most of them are redundant. It is mostly a way of ensuring the pre-commit hook itself is valid (second-order QA :smile: )

chifflier:
> Seems OK if you like pre-commit hooks :-)
> 
> I had some nits to consider but nothing I consider blocking. Thanks!

Thanks for the review, I will push some updates then merge.
For the record: I usually do not really like pre-commit hooks, but I had several projects where I found them useful to catch errors before the CI. I think of these hooks as bonus (and optional)

.github/workflows/rust.yml
cpu: That makes sense and I don't think it'll blow up CI time :-) Happy to have it stay as-is. I'll resolve the thread.

